PR #40, opened 2018-10-18
Add Windows Integrated Auth when possible. Default to long-lived PATs instead of JWTs when interactive is required.

zarenner:


zarenner:
@jmyersmsft @satbai @satishkumaarponnusamy

CredentialProvider.Microsoft/Util/EnvUtil.cs
```diff
@@ -21,6 +22,7 @@ public static class EnvUtil
 
         public const string SupportedHostsEnvVar = "NUGET_CREDENTIALPROVIDER_VSTS_HOSTS";
         public const string SessionTimeEnvVar = "NUGET_CREDENTIALPROVIDER_VSTS_SESSIONTIMEMINUTES";
+        public const string TokenTypeEnvVar = "NUGET_CREDENTIALPROVIDER_VSTS_TOKENTYPE";
```
satbai: Please add this env var to the EnvironmentVariableHelp text in Resources file. Also makes sense to update it in the readme

zarenner:
Thanks for the review @satbai. I don't have contributor permission, so (assuming there aren't other issues), could you please merge this?

CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsCredentialProvider.cs
```diff
@@ -81,8 +78,35 @@ public override async Task<GetAuthenticationCredentialsResponse> HandleRequestAs
                         responseCode: MessageResponseCode.Error);
                 }
 
-                var sessionTokenClient = new VstsSessionTokenClient(request.Uri, bearerToken, authUtil);
-                var sessionToken = await sessionTokenClient.CreateSessionTokenAsync(endTime, cancellationToken);
+                // Allow the user to choose their token type
+                // If they don't and interactive auth was required, then prefer a PAT so we can safely default to a much longer validity period
+                VstsTokenType tokenType = EnvUtil.GetVstsTokenType() ??
+                    (bearerTokenResult.ObtainedInteractively
```
johnterickson, line 84: @satbai or @zarenner  - Any recollection on why this pivoted on interactive vs non-interactive?

jmyersmsft: @johnterickson because if we could obtain the token non-interactively, we assume we could obtain the token again non-interactively, so it's safe to have a very short validity period without annoying the user with re-auths, and if we have a very short validity period, we can just request a self-describing token that's a tiny bit more efficient on the server and doesn't clutter up the user's PAT UI

If we had to prompt the user, then we wanted to request a longer lifetime so we don't annoy them with repeated prompts, for which we wanted to obtain a PAT, which shows up in the UI and can be revoked or extended by the user as usual for PATs

PR #50, opened 2019-01-10
Add more verbose logging and move stuff around

satbai:
- Add more logging
- Only try cache if cred provider is cachable
- Remove IsRetry check from build task cred providers
- In build task cred providers, move all checks but environment variable checks from CanProvideCredentials to HandleRequest
- Only use VstsCredentialProvider if build task env vars are not set
- Make the actionable device flow console message easier to see
image
- Add special cancellation handling to fix a "message type 'Cancel' is invalid at this time" issue

### DIFF
--- a/CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsCredentialProvider.cs
+++ b/CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsCredentialProvider.cs
@@ -37,20 +37,6 @@ namespace NuGetCredentialProvider.CredentialProviders.Vsts
 
         public override async Task<bool> CanProvideCredentialsAsync(Uri uri)
         {
-            string buildTaskJsonEnvVar = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskExternalEndpoints);
-            string buildTaskPrefixesEnvVar = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskUriPrefixes);
-            string buildTaskAccessTokenEnvVar = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskAccessToken);
-
-            bool shouldUseBuildTaskCredProvider = string.IsNullOrWhiteSpace(buildTaskJsonEnvVar) == false 
-                || string.IsNullOrWhiteSpace(buildTaskPrefixesEnvVar) == false 
-                || string.IsNullOrWhiteSpace(buildTaskAccessTokenEnvVar) == false;
-            if (shouldUseBuildTaskCredProvider)
-            {
-                // If env vars for build task cred providers are set, this cred provider cannot be used.
-                Verbose(Resources.BuildTaskCredProviderIsUsedError);
-                return await Task.FromResult(false);
-            }
-
             var validHosts = EnvUtil.GetHostsFromEnvironment(Logger, EnvUtil.SupportedHostsEnvVar, new[]
             {
                 ".pkgs.vsts.me", // DevFabric
--- a/CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsCredentialProvider.cs
+++ b/CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsCredentialProvider.cs
@@ -37,6 +37,20 @@ namespace NuGetCredentialProvider.CredentialProviders.Vsts
 
         public override async Task<bool> CanProvideCredentialsAsync(Uri uri)
         {
+            string buildTaskJsonEnvVar = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskExternalEndpoints);
+            string buildTaskPrefixesEnvVar = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskUriPrefixes);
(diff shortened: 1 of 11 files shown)


CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsCredentialProvider.cs
```diff
@@ -37,6 +37,20 @@ public sealed class VstsCredentialProvider : CredentialProviderBase
...
         public override async Task<bool> CanProvideCredentialsAsync(Uri uri)
         {
+            string buildTaskJsonEnvVar = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskExternalEndpoints);
+            string buildTaskPrefixesEnvVar = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskUriPrefixes);
+            string buildTaskAccessTokenEnvVar = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskAccessToken);
+
+            bool shouldUseBuildTaskCredProvider = string.IsNullOrWhiteSpace(buildTaskJsonEnvVar) == false 
+                || string.IsNullOrWhiteSpace(buildTaskPrefixesEnvVar) == false 
+                || string.IsNullOrWhiteSpace(buildTaskAccessTokenEnvVar) == false;
+            if (shouldUseBuildTaskCredProvider)
```
shubham90: I thought the control first goes to build task cred provider. If it has already handled the request, why would the request reach this stage?

satbai: There were cases where CanProvideCredentials check would return false from the build task cred providers even when the build task environment variables were set, and I don't think this cred provider should be tried even as a backup to avoid confusion in the possible error logs. As an alternative, I could only check the env vars in the build task CanProvideCredentials methods and move the other checks to the HandleRequestAsync method. I think I like that more - I'll make the change. This snippet here might just confuse people. Thanks!

satbai: done in commit https://github.com/Microsoft/artifacts-credprovider/pull/50/commits/7d44855eb772c504aadb54d8eab8151389de4ec8

CredentialProvider.Microsoft/CredentialProviders/VstsBuildTask/VstsBuildTaskCredentialProvider.cs
```diff
@@ -63,7 +46,30 @@ public override Task<GetAuthenticationCredentialsResponse> HandleRequestAsync(Ge
         {
             cancellationToken.ThrowIfCancellationRequested();
 
+            string uriPrefixesString = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskUriPrefixes);
             string accessToken = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskAccessToken);
+
+            string[] uriPrefixes = uriPrefixesString.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
```
shubham90: Can we add a null or empty check on uriPrefixesString here?

CredentialProvider.Microsoft/RequestHandlers/GetAuthenticationCredentialsRequestHandler.cs
```diff
@@ -78,6 +67,16 @@ public override async Task<GetAuthenticationCredentialsResponse> HandleRequestAs
                     continue;
                 }
 
+                if (credentialProvider.IsCachable && TryCache(request, out string cachedToken))
```
shubham90: We can store the output for TryCache() outside the for loop and use the result inside. Saves us some calls to cache.

shubham90:
Left couple of comments. Otherwise, looks good.

CredentialProvider.Microsoft/RequestHandlers/GetAuthenticationCredentialsRequestHandler.cs
satbai: We only want to try cache if the cred provider is cachable and we need to check it inside the loop. Currently only 1 out of 3 is cachable. With build task cred providers we never want to try cache.

CredentialProvider.Microsoft/CredentialProviders/VstsBuildTask/VstsBuildTaskCredentialProvider.cs
satbai: There is a null check in the CanProvideCredentials method so no need to do it again here

CredentialProvider.Microsoft/RequestHandlers/GetAuthenticationCredentialsRequestHandler.cs
shubham90: ah ok. makes sense then.

ViktorHofer:
Quick question as I'm a bit confused about how we should use env variables and the output logging after this PR. Should we set env variables when using the credential provider or should we not set such?

In corefx we currently frequently hit 401 errors. This is the piece where we restore with the credentials provider integrated in nuget.exe: https://github.com/dotnet/corefx/blob/master/eng/pipelines/corefx-base.yml#L133-L144

PR #52, opened 2019-01-12
Adding netcore outputs to cred provider

shubham90:


build/build.yml
```diff
@@ -101,6 +94,32 @@ steps:
     targetFolder: '$(Build.ArtifactStagingDirectory)\tarball\'
     flattenFolders: true
 
+- task: ArchiveFiles@2
+  displayName: Create netcore tarball
+  inputs:
+    rootFolderOrFile: '$(Build.ArtifactStagingDirectory)\tarball\'
```
zarenner: Not part of your change, but a bit confusing IMO that the 'tarball' folder is used to produce both a tarball/tgz, and a zip.

build/build.yml
```diff
@@ -101,6 +94,32 @@ steps:
...
+    includeRootFolder: false
+    archiveType: 'zip'
+    archiveFile: '$(Build.ArtifactStagingDirectory)\$(BuildConfiguration)\Microsoft.NetCore.NuGet.CredentialProvider.zip'
+    replaceExistingArchive: true
+
+- task: CopyFiles@2
+  displayName: Copy netfx files into tarball
+  inputs:
+    sourceFolder: '$(Build.SourcesDirectory)\CredentialProvider.Microsoft\bin\$(BuildConfiguration)\net461\'
```
zarenner: net461 hardcoded? I assume it'll at least break the build if we upgraded netfx ver?

build/build.yml
shubham90: yeah, I saw that too, forgot to make the change with this PR though. Will change in a separate PR.

build/build.yml
shubham90: good catch! will fix this as weel in follow up PR.

PR #58, opened 2019-01-17
Fix PowerShell script

satbai:


build/build.yml
```diff
@@ -1,165 +1,167 @@
 parameters:
```
satbai: My VSCode just added some styling I didn't want - will undo

Issue #46, opened 2018-12-26
Corrupted 'SessionTokenCache.dat' produces inactionable, unrecoverable 'The data is invalid' errors


bergbria:
## Overview
I installed the cred provider to interact with an Azure DevOps packages feed. All was well.

At some point (possibly after a Blue Screen, though I'm not sure), all of my nuget commands started failing with the following error message.
```Terminating plugin 'CredentialProvider.Microsoft' due to an unrecoverable fault:  NuGet.Protocol.Plugins.ProtocolException: A plugin protocol exception occurred. ---> NuGet.Protocol.Plugins.ProtocolException: The data is invalid.

   at NuGet.Protocol.Plugins.OutboundRequestContext`1.HandleFault(Message fault)
   at NuGet.Protocol.Plugins.MessageDispatcher.OnMessageReceived(Object sender, MessageEventArgs e)
   at System.EventHandler`1.Invoke(Object sender, TEventArgs e)
   at NuGet.Protocol.Plugins.StandardOutputReceiver.OnLineRead(Object sender, LineReadEventArgs e)
   --- End of inner exception stack trace ---
```

The error didn't go away after re-runs and a reboot.
The errors provided no hint as to what the issue was or how it could be resolved.

After debugging for a while, I was able to trace the error to the following call stack:
```>	System.Security.dll!System.Security.Cryptography.ProtectedData.Unprotect(byte[], byte[], System.Security.Cryptography.DataProtectionScope)	Unknown
 	credentialprovider.microsoft.exe!NuGetCredentialProvider.Util.EncryptedFile.ReadFileBytes(string, bool) Line 17	C#
 	credentialprovider.microsoft.exe!NuGetCredentialProvider.Util.SessionTokenCache.ReadFileBytes() Line 85	C#
 	credentialprovider.microsoft.exe!NuGetCredentialProvider.Util.SessionTokenCache.Cache.get() Line 27	C#
 	credentialprovider.microsoft.exe!NuGetCredentialProvider.Util.SessionTokenCache.TryGetValue(System.Uri, out string) Line 53	C#
 	credentialprovider.microsoft.exe!NuGetCredentialProvider.RequestHandlers.GetAuthenticationCredentialsRequestHandler.TryCache(NuGet.Protocol.Plugins.GetAuthenticationCredentialsRequest, out string) Line 149	C#
[...]

PR #59, opened 2019-01-18
Catch an exception when reading cache

satbai:
Fixes [#46](https://github.com/Microsoft/artifacts-credprovider/issues/46) by removing the cached token on exception.

### DIFF
--- a/CredentialProvider.Microsoft/Resources.resx
+++ b/CredentialProvider.Microsoft/Resources.resx
@@ -416,6 +416,6 @@ Device Flow Authentication Timeout
     <value>Caught exception processing {0}  (RequestId: {1})</value>
   </data>
   <data name="CacheException" xml:space="preserve">
-    <value>Exception happened when trying to get cached token: invalidating cache. Message: {0}</value>
+    <value>Exception happened when trying to get cached token. Invalidating cache. Message: {0}</value>
   </data>
 </root>
--- a/CredentialProvider.Microsoft/Util/SessionTokenCache.cs
+++ b/CredentialProvider.Microsoft/Util/SessionTokenCache.cs
@@ -14,9 +14,12 @@ namespace NuGetCredentialProvider.Util
     {
         private static readonly object FileLock = new object();
         private readonly string cacheFilePath;
-        public SessionTokenCache(string cacheFilePath)
+        private ILogger logger;
+
+        public SessionTokenCache(string cacheFilePath, ILogger logger)
         {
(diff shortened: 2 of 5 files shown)


CredentialProvider.Microsoft/Resources.resx
```diff
@@ -415,4 +415,7 @@ Device Flow Authentication Timeout
   <data name="ResponseHandlerException" xml:space="preserve">
     <value>Caught exception processing {0}  (RequestId: {1})</value>
   </data>
+  <data name="CacheException" xml:space="preserve">
+    <value>Exception happened when trying to get cached token: invalidating cache. Message: {0}</value>
```
shubham90: nit: "cached token. Invalidating cache. Message: {0} "

CredentialProvider.Microsoft/Util/SessionTokenCache.cs
```diff
@@ -4,15 +4,16 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using Newtonsoft.Json;
+using NuGetCredentialProvider.Logging;
 
 namespace NuGetCredentialProvider.Util
 {
     public class SessionTokenCache : ICache<Uri, string>
```
shubham90: Can we add logger to this class as a variable and not in the interface method? TryGetValue(Uri key, out string value) looks more correct than having to pass a logger.

CredentialProvider.Microsoft/Util/SessionTokenCache.cs
```diff
@@ -48,9 +49,26 @@ public bool ContainsKey(Uri key)
...
+                return Cache.TryGetValue(key, out value);
+            }
+            catch (Exception e)
+            {
+                if (File.Exists(cacheFilePath))
+                {
+                    logger.Verbose(string.Format(Resources.CacheException, e.Message));
+
+                    File.Delete(cacheFilePath);
+                    Cache.Remove(key);
```
shubham90: If the cache is corrupted, we should reset the entire Cache object rather than removing just one key.

CredentialProvider.Microsoft/Resources.resx
satbai: Thanks! Done

CredentialProvider.Microsoft/Util/SessionTokenCache.cs
satbai: Good idea - done!

Issue #60, opened 2019-01-21
Self-hosted build agent and dotnet restore


danielwertheim:
How are we supposed to configure this if a self-hosted agent is installed on a own server and a build script calls dotnet restore? I have a custom nuget.config that clears and adds:

1 - a Azure DevOps Artifacts feed
2 - NuGet org

Is it supposed to work if I:

Add environment variable: `NUGET_PLUGIN_PATHS` pointing to folder with the ZIPs plugins

Add environment variable: `VSS_NUGET_EXTERNAL_FEED_ENDPOINTS` with JSON

```
{"endpointCredentials": [{"endpoint":"https://myfeed.com/index.json", "username":"not-required", "password":"accesstoken"}]}
```

The access token has full rights to Packages feeds.

zjrunner:
Is your self-hosted agent an Azure Devops Pipelines agent?  I'm trying to understand if you're running in Pipelines, but just using a manual script to restore instead of the existing pipeline tasks, or if you're trying to run in a separate CI/CD ecosystem entirely.

satbai:
If you're using the NUGET_PLUGIN_PATHS variable, please make sure that it points to the .exe with NuGet or .dll with dotnet instead of the folder. For example, `NUGET_PLUGIN_PATHS="C:\path-to-cred-provider\CredentialProvider.Microsoft.dll"`. Also, instead of using the environment variable, you can use the [script](https://github.com/Microsoft/artifacts-credprovider/blob/master/helpers) to install the cred provider to the location that NuGet and dotnet knows to look for the plugin.

However, please let us know if you're using the pipeline tasks mentioned above - in that case I'll have to look into the task you're using.

danielwertheim:
@zjrunner  Using Azure DevOps: 
- Repo
- Pipeline
- Artifacts feed

Using a devops pipeline agent. Everything works, except authentication against the Azure DevOps artifacts feed when it comes to restoring via `dotnet build` in the build script (CakeBuild in via Powershell task).

satbai:
We recommend doing dotnet restore as a separate step before you do dotnet build to make it easier to trouble shoot. If dotnet restore still fails after changing the NUGET_PLUGIN_PATHS variable to point to the .dll and verifying you're using dotnet version 2.1.400 or greater, please run it with detailed verbosity and share the logs so we can investigate further.

danielwertheim:
[...]

PR #61, opened 2019-01-23
Update README.md

danielwertheim:
Adds tip about possible necessary restart after adding environment variables.

Which was what caused issues for me in issue #60


alexmullans:
Thanks for the contribution!

PR #66, opened 2019-02-14
Simplify check for Security protocol to not rely on string manipulation

bergmeister:


jotaylo:
This doesn't work as expected for me

```
PS D:\> [Net.ServicePointManager]::SecurityProtocol
Tls, Tls11, Tls12

PS D:\> [Net.ServicePointManager]::SecurityProtocol -notcontains [System.Net.SecurityProtocolType]::Tls12
True
```

Since SecurityProtocol is basically a flagged enum, looking for the flag does the trick

```
PS D:\> ([Net.ServicePointManager]::SecurityProtocol -band [System.Net.SecurityProtocolType]::Tls12) -ne [System.Net.SecurityProtocolType]::Tls12
False
```

Or, could just always add it
```
[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [System.Net.SecurityProtocolType]::Tls12
```

PR #70, opened 2019-02-22
Add stopwatch logging to request handler

satbai:
-Adding stopwatch logging to request handler
-Removing custom cancel call to nuget


CredentialProvider.Microsoft/RequestHandlers/RequestHandlerBase.cs
```diff
@@ -65,20 +68,13 @@ public async Task HandleResponseAsync(IConnection connection, Message message, I
                 catch (Exception ex) when (cancellationToken.IsCancellationRequested)
                 {
                     Logger.Verbose(string.Format(Resources.RequestHandlerCancelingExceptionMessage, ex.InnerException, ex.Message));
-                    Logger.Verbose(Resources.CancellationRequested);
-
```
shubham90: Can you please add a comment here stating why we are not sending cancellation message to nuget? Maybe something like: Nuget handles the cancellation event and eventually times out because in some scenarios it cannot handle the cancellation message properly.

PR #82, opened 2019-03-25
Make Linux/Mac install helper more robust

aude:
The install script does not work by default at my Arch Linux system. It assumes `python` means Python 2, but it means Python 3 at some systems. This pull request features a bug fix for this.

After looking at it a bit, I noticed that several shell script best practices are not implemented. As the shell script enthusiast I am, I thought I'd contribute a couple improvements :smile: Mainly suggestions from the tool [shellcheck](https://www.shellcheck.net/).

### DIFF
--- a/helpers/installcredprovider.sh
+++ b/helpers/installcredprovider.sh
@@ -14,7 +14,7 @@ NUGET_PLUGIN_DIR="$HOME/.nuget/plugins"
 # Ensure plugin directory exists
 if [[ ! -e ${NUGET_PLUGIN_DIR} ]]; then
   echo "INFO: Creating the nuget plugin directory (i.e. ${NUGET_PLUGIN_DIR}). "
-  if ! mkdir -p ${NUGET_PLUGIN_DIR}; then
+  if ! mkdir -p "${NUGET_PLUGIN_DIR}"; then
       echo "ERROR: Unable to create nuget plugins directory (i.e. ${NUGET_PLUGIN_DIR})."
       exit 1
   fi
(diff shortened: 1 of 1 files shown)


jmyersmsft:
I'm going to close this in favor of #83 which removes the Python dependency entirely. I've incorporated the fixes from this PR that are still relevant, thanks!

aude:
This was a bit demotivating for me.

If my code is accepted I would love to have an entry in the commit history. That's the biggest reward for me personally, when I'm being a volunteer contributor. Without that, I feel a bit empty to be honest.

Anyway, nothing to do about it now. Just thought I'd be open here. Maybe you'll motivate more contributors by doing things a tiny bit differently. One example could be to accept pull requests and doing modifications afterwards.

jmyersmsft:
Hmm, I tried to give you at least some credit by mentioning you in the PR description, but I didn't realize that unlike Azure Repos, GitHub doesn't copy that to the merge commit message. 

I didn't take your PR directly because we were concerned that while it's true that `python` might be either Python 2 or Python 3, `python3` also isn't guaranteed to be available. 

In this small repo, it should be easy to just fix this attribution problem, so I'd like to do that. If you agree, I'll revert my PR, then merge a new one that just removes the Python dependency, then you can reopen this PR, resolve the conflict with my change, and I'll merge it in. That'll properly represent the authorship without taking the part of your PR that we found problematic.

Thank you again for your contribution!

aude:
I appreciate it, thanks for understanding!

I realize it might be a bit shallow to wish for credit like this, it makes no difference to the code itself. But I think it's helpful to be open about what drives me to contribute.

Your idea sounds super to me, go ahead.

jmyersmsft:
#87 is in, go ahead and update this PR.

aude:
Updated.

Issue #84, opened 2019-04-03
installcredprovider.sh errors on WSL (Ubuntu)


ViktorHofer:
/home/vihofer/corefx/artifacts/installcredprovider.sh: 16: /home/vihofer/corefx/artifacts/installcredprovider.sh: [[: not found

Seems like https://github.com/Microsoft/artifacts-credprovider/blob/master/helpers/installcredprovider.sh#L16 is not supported on all bash variants?

PR #85, opened 2019-04-03
Use cross-plat dir check in installcredprovider.sh

ViktorHofer:
Fixes https://github.com/Microsoft/artifacts-credprovider/issues/84

See https://stackoverflow.com/questions/59838/check-if-a-directory-exists-in-a-shell-script#comment70582468_59839

### DIFF
--- a/helpers/installcredprovider.sh
+++ b/helpers/installcredprovider.sh
@@ -13,7 +13,7 @@ RELEASEURL="$GITHUB/$REPO/releases/$VERSION/download/$FILE"
 NUGET_PLUGIN_DIR="$HOME/.nuget/plugins"
 
 # Ensure plugin directory exists
-if [[ ! -e "${NUGET_PLUGIN_DIR}" ]]; then
+if [ ! -d "${NUGET_PLUGIN_DIR/}" ]; then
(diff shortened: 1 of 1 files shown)


helpers/installcredprovider.sh
```diff
@@ -13,7 +13,7 @@ RELEASEURL="$GITHUB/$REPO/releases/$VERSION/download/$FILE"
 NUGET_PLUGIN_DIR="$HOME/.nuget/plugins"
 
 # Ensure plugin directory exists
-if [[ ! -e "${NUGET_PLUGIN_DIR}" ]]; then
+if [ ! -d "${NUGET_PLUGIN_DIR/}" ]; then
```
jmyersmsft: This slash isn't quite doing what the linked SO question was suggesting. It may even be trying to include the / in the name of the variable. We don't really care if the dir is a symlink in this case, so the extra / is unnecessary. If it's a symlink to a regular file, mkdir will fail, but that should be pretty rare and self-explanatory.

```suggestion
if [ ! -d "${NUGET_PLUGIN_DIR}" ]; then
```

jmyersmsft:
You'll want to hold for #82 before trying to resolve conflicts, as it'll probably resolve the conflict

ViktorHofer:
done.

jmyersmsft:
Thank you for your contribution!

PR #95, opened 2019-05-23
Remove instructions not to install netfx

zarenner:
Installing the netfx plugin is required to support nuget.exe, as the plugin installed by VS is not currently discovered by nuget.exe (only msbuild /restore)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,9 +34,7 @@ If you are using `dotnet` or `nuget`, you can use the Azure Artifact Credential 
 
 1. Download the latest release of [Microsoft.NuGet.CredentialProvider.zip](https://github.com/Microsoft/artifacts-credprovider/releases)
 2. Unzip the file
-3. Copy the `netcore` directory from the extracted archive to `$env:UserProfile\.nuget\plugins`
-
-Note: copying the `netfx` directory is not recommended unless you don't plan to install an edition of Visual Studio (including the Build Tools edition). All Visual Studio editions come with (and automatically update) the `netfx` version of the Credential Provider.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,9 +34,7 @@ If you are using `dotnet` or `nuget`, you can use the Azure Artifact Credential
 
 1. Download the latest release of [Microsoft.NuGet.CredentialProvider.zip](https://github.com/Microsoft/artifacts-credprovider/releases)
 2. Unzip the file
-3. Copy the `netcore` directory from the extracted archive to `$env:UserProfile\.nuget\plugins`
-
-Note: copying the `netfx` directory is not recommended unless you don't plan to install an edition of Visual Studio (including the Build Tools edition). All Visual Studio editions come with (and automatically update) the `netfx` version of the Credential Provider.
+3. Copy the `netcore` (and optionally `netfx` for nuget.exe) directory from the extracted archive to `$env:UserProfile\.nuget\plugins`
```
infin8x: Hmm, OK. Now I'm remembering something about a good reason for having this warning. @anangaur can you shed some light on why we may have not wanted to install the netfx version of the credprovider into the NuGet plugins folder? I'm vaguely remembering something about that taking precedence of the version that comes with VS and thus never getting updated.

zarenner: Yes, that is definitely a concern (we were just discussing this in the team room), and we need to solve that as part of the longer-term acquisition story. IMO nuget should somehow choose the latest/most preferred or something like that and make it clear in logs which is being used.

If @anangaur has ideas for easily having nuget.exe use the plugin with VS, that would probably be preferable to avoid this issue. But today nuget.exe doesn't "just work" after installing VS and the current doc text implies it does.

infin8x: Makes sense. I like the current wording - let's run with that.

PR #98, opened 2019-06-05
Log to stderr (instead of not logging) when emitting the result as JSON

zarenner:
Device code mechanism needs to emit logs to the console, but in JSON mode we cannot emit logs to stdout since we need to keep the output pure JSON.


CredentialProvider.Microsoft/Logging/ConsoleLoggers.cs
```diff
@@ -0,0 +1,44 @@
...
+    internal class HumanFriendlyTextWriterLogger : LoggerBase
+    {
+        private readonly TextWriter writer;
+
+        public HumanFriendlyTextWriterLogger(TextWriter writer)
+        {
+            this.writer = writer;
+        }
+
+        protected override void WriteLog(LogLevel logLevel, string message)
```
zarenner: The base class is what adds on [CredentialProvider] to the start of {message}. Should probably re-factor at some point since that seems odd to me, but didn't want to touch right now.

CredentialProvider.Microsoft/Logging/ConsoleLoggers.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// </summary>
+    internal class StandardErrorLogger : HumanFriendlyTextWriterLogger
+    {
+        public StandardErrorLogger() : base(Console.Error) { }
+    }
+
+    /// <summary>
+    /// Emits a log message in a human-readable format to a TextWriter, including the log level
+    /// </summary>
+    internal class HumanFriendlyTextWriterLogger : LoggerBase
```
zarenner: FYI FileLogger could also use this (currently it does its own formatting), but didn't want to make that change yet.

Issue #89, opened 2019-04-04
installcredprovider script should respect NUGET_PACKAGES env variable


ViktorHofer:
See https://docs.microsoft.com/en-us/nuget/consume-packages/managing-the-global-packages-and-cache-folders

satbai:
Hi @ViktorHofer, I just checked the NUGET_PACKAGES variable and tried it out. It looks like for NuGet to find the auth plugin, the cred provider still needs to live in %userprofile%/.nuget/plugins even if I use the  the NUGET_PACKAGES variable and the packages folder isn't %userprofile%/.nuget/packages anymore. There is a chance that I misunderstood the ask, however. Please feel free to re-open the issue if I misunderstood, and explain in detail what the request for the NUGET_PACKAGES env var is. Thanks!

ViktorHofer:
The script currently hardcodes the path to the packages folder and does not honor the NUGET_PACKAGES environment variable: https://github.com/microsoft/artifacts-credprovider/blob/master/helpers/installcredprovider.ps1#L23

satbai:
That is the location of the plugin, not the packages folder. Was the install script failing for you? If it was, maybe the .nuget folder didn't exist? If that's the case I can modify the script to check for the folder.

ViktorHofer:
Right, why does the plugin need to be installed at that location? Why does nuget not try to resolve the same way as it determines the packages folder location?

satbai:
Here's the [spec ](https://github.com/NuGet/Home/wiki/NuGet-Cross-Plat-Credential-Plugin#plugin-installation-and-discovery)on how NuGet finds the cred provider plugin. Basically, %userprofile%/.nuget/plugins is a location NuGet has chosen to look from when trying to authenticate to a package source endpoint. That's why the script always installs the plugin to the same location.

ViktorHofer:
Can we change that / is there a reason for that?

satbai:
I don't think the location can currently be changed by users. Although, if you need to have the cred provider in a different location, you can manually move it to a location you prefer and then use the NUGET_PLUGIN_PATHS environment variable described in the spec to point to the cred provider .exe or .dll file.

satbai:
What we could do is add a comment to the script or update the readme and explain that if the cred provider needs to be in a different location for some reason, you can use the NUGET_PLUGIN_PATHS variable. Would that have been helpful?

ViktorHofer:
[...]

PR #109, opened 2019-07-12
Update readme

satbai:
Making the readme better
https://github.com/microsoft/artifacts-credprovider/issues/110 
https://github.com/microsoft/artifacts-credprovider/issues/89

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,23 +28,23 @@ Install [Visual Studio version 15.9-preview1 or later](https://visualstudio.micr
 
 ## Setup
 
-If you are using `dotnet` or `nuget`, you can use the Azure Artifact Credential Provider by adding it to [NuGet's plugin search path](https://github.com/NuGet/Home/wiki/NuGet-Cross-Plat-Credential-Plugin#plugin-installation-and-discovery). This section contains both manual and scripted instructions for doing so. 
+If you are using `dotnet` or `nuget`, you can use the Azure Artifact Credential Provider by adding it to [NuGet's plugin search path](https://github.com/NuGet/Home/wiki/NuGet-Cross-Plat-Credential-Plugin#plugin-installation-and-discovery). This section contains both manual and scripted instructions for doing so.
 
-### Installation on Windows:
+### Installation on Windows
 
-#### Automatic PowerShell script:
+#### Automatic PowerShell script
 
 [PowerShell helper script](helpers/installcredprovider.ps1)
 - To install netcore, run `installcredprovider.ps1`
 - To install both netfx and netcore, run `installcredprovider.ps1 -AddNetfx`. The netfx version is needed for nuget.exe.
 
-### Manual installation on Windows:
+#### Manual installation on Windows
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,27 +30,35 @@ Install [Visual Studio version 15.9-preview1 or later](https://visualstudio.micr
...
+
+#### Manual installation on Windows
 
 1. Download the latest release of [Microsoft.NuGet.CredentialProvider.zip](https://github.com/Microsoft/artifacts-credprovider/releases)
 2. Unzip the file
-3. Copy the `netcore` (and optionally `netfx` for nuget.exe) directory from the extracted archive to `$env:UserProfile\.nuget\plugins`
+3. Copy the `netcore` (and `netfx` for nuget.exe) directory from the extracted archive to `$env:UserProfile\.nuget\plugins`
 
-### Automatic PowerShell installation:
+Using the above is recommended, but as per [NuGet's plugin discovery rules](https://github.com/NuGet/Home/wiki/NuGet-Cross-Plat-Credential-Plugin#plugin-installation-and-discovery), alternatively you can install the credential provider to a location you prefer, and then set the environment variable NUGET_PLUGIN_PATHS to the .exe of the credential provider found in plugins\netfx\CredentialProvider.Microsoft\CredentialProvider.Microsoft.exe. For example, $env:NUGET_PLUGIN_PATHS="my-alternative-location\CredentialProvider.Microsoft.exe".
```
zarenner: This will break `dotnet` (at least until https://github.com/NuGet/Home/issues/8151 is fixed). Please update this guidance to make that clear.

satbai: Thanks! I'll remove it from under dotnet and leave it for under nuget

satbai: Hmm, nuget has the issue as well? Maybe I'll just add a link to the bug here. I'll think about this one.

zarenner: The "Linux and Mac" instructions below are perfectly fine, since dotnet is the only possibility there.

You don't need to remove it from here IMO, it should just indicate that setting the envvar to the .exe credprovider will break dotnet on windows, and setting the envvar to the .dll will break nuget.exe and MSBuild on windows.

zarenner: Therefore customers that use both dotnet and nuget.exe/MSBuild (.NET Framework) should probably avoid that envvar for now.

satbai: I added "Note that if you are using both nuget and dotnet, this environment variable is not recommended due to this issue: https://github.com/NuGet/Home/issues/8151"

zarenner: Looks good for the windows section, but is unnecessary for the linux/mac section (running nuget.exe and the .exe credprovider is not supported)

README.md
```diff
@@ -88,6 +96,12 @@ msbuild /t:restore /p:nugetInteractive=true
 
 Once you've successfully acquired a token, you can run authenticated commands without the `/p:nugetInteractive=true` switch.
 
+### unattended build agents
```
zarenner: Nit: capitalize

README.md
satbai: Thanks. I guess I'm ready for the weekend.

Issue #108, opened 2019-07-11
'dotnet restore' unhandled exception with NuGetCredentialProvider


AMoghrabi:
Hey team,

I'm not sure if this is an issue with the dotnet CLI or with the credential provider. I'll try this repository first. I've also reviewed the other open issues and it this doesn't appear to be related.

We are using Azure Artifacts and restoring packages in a docker container:

```
WORKDIR /nuget_cache_warmer
COPY **/*.csproj ./
RUN for f in ./*.csproj; do dotnet restore $f --source $NUGET_SOURCE --source https://api.nuget.org/v3/index.jsonl; done
```

This is our base image: `FROM microsoft/dotnet:2.1-sdk`

Intermittently, I start to see: 

```
Unhandled Exception:  [0m [91mSystem.Threading.Tasks.TaskCanceledException: A task was canceled.
NuGet.Protocol.Plugins.MessageDispatcher.DispatchWithNewContextAsync[TOutgoing,TIncoming](IConnection connection, MessageType type, MessageMethod method, TOutgoing payload, CancellationToken cancellationToken) at NuGet.Protocol.Plugins.SymmetricHandshake.HandshakeAsync(CancellationToken cancellationToken) at NuGet.Protocol.Plugins.Connection.ConnectAsync(CancellationToken cancellationToken)
at NuGet.Protocol.Plugins.PluginFactory.CreateFromCurrentProcessAsync(IRequestHandlers requestHandlers, ConnectionOptions options, CancellationToken sessionCancellationToken)
at NuGetCredentialProvider.Program.Main(String[] args)
at NuGetCredentialProvider.Program.<Main>(String[] args)
```

This happens more than once during the restore processing, but not for every `.csproj` file. 

What is interesting compared to other issues is the restore process will complete, and I can run the next docker step to build my project successfully. 

If it matters, the docker build is running on a Linux TeamCity agent. If I run `dotnet restore --disable-parallel`, the exception does not occur, however the restore takes twice as long (as expected).

Since I can still restore successfully this isn't a huge problem, though I thought I'd report it.

satbai:
This could be a docker issue or related to this open issue with dotnet https://github.com/NuGet/Home/issues/7842.
FYI @shubham90 who worked on the docker implementation - I forgot if the TaskCanceledException was resolved with some of the docker images?

AMoghrabi:
[...]

PR #115, opened 2019-07-31
Fix Unhandled TaskCanceledException

AMoghrabi:
Hey team,

This is a fix for my issue  https://github.com/microsoft/artifacts-credprovider/issues/108

I've taken @dtivel's suggestion of wrapping a block of code in `Program.cs` to catch an unhandled `TaskCanceledException`. 

I ran the unit tests in the project and all but two tests pass. The two that fail are:
```
CanProvideCredentials_ReturnsTrueForKnownSources()
CanProvideCredentials_ReturnsTrueForOverridenSources()
```

However, these tests also fail when running against the latest in `master`. 

As far as I can tell, my issue can only be replicated in a docker container. I've ran some before and after tests using the latest version of the cred provider, and a copy of the cred provider after I have made my change. I no longer receive the error:

```
Unhandled Exception: System.Threading.Tasks.TaskCanceledException: A task was canceled.`
```

However, I am now seeing about 5 warnings similar to this:

```
The HTTP request to 'GET https://pkgs.dev.azure.com/{company}/_packaging/47a1753d-0296-432f-95f2-2f9e908f3e8f/nuget/v3/flat2/microsoft.aspnetcore.http.abstractions/index.json' has timed out after 100000ms.
```

But it doesn't appear to be causing the build to fail. I also want to mention that the error is likely because we don't have upstream sources enabled, and it's likely a `dotnet restore` bug causing the timeouts. 

Looking forward to any advice to improve my submission and having this merged & released so we can move forward with Azure Artifacts 😄 

Thanks!

### DIFF
--- a/CredentialProvider.Microsoft.Tests/CredentialProviders/VstsBuildTaskServiceEndpoint/VstsBuildTaskServiceEndpointCredentialProviderTests.cs
+++ b/CredentialProvider.Microsoft.Tests/CredentialProviders/VstsBuildTaskServiceEndpoint/VstsBuildTaskServiceEndpointCredentialProviderTests.cs
@@ -126,5 +126,21 @@ namespace CredentialProvider.Microsoft.Tests.CredentialProviders.VstsBuildTaskSe
             Func<Task> act = async () => await vstsCredentialProvider.HandleRequestAsync(new GetAuthenticationCredentialsRequest(sourceUri, false, false, false), CancellationToken.None);
             act.Should().Throw<Exception>();
         }
+
+        [TestMethod]
+        public async Task HandleRequestAsync_MatchesEndpointURLCaseInsensitive()
+        {
+            Uri sourceUri = new Uri(@"http://example.pkgs.vsts.me/_Packaging/TestFEED/nuget/v3/index.json");
+
+            string feedEndPointJson = "{\"endpointCredentials\":[{\"endpoint\":\"http://example.pkgs.vsts.me/_packaging/TestFeed/nuget/v3/index.json\", \"username\": \"testUser\", \"password\":\"testToken\"}]}";
+            string feedEndPointJsonEnvVar = EnvUtil.BuildTaskExternalEndpoints;
+
+            Environment.SetEnvironmentVariable(feedEndPointJsonEnvVar, feedEndPointJson);
+
+            var result = await vstsCredentialProvider.HandleRequestAsync(new GetAuthenticationCredentialsRequest(sourceUri, false, false, false), CancellationToken.None);
+            Assert.AreEqual(MessageResponseCode.Success, result.ResponseCode);
+            Assert.AreEqual("testUser", result.Username);
+            Assert.AreEqual("testToken", result.Password);
+        }
     }
 }
--- a/CredentialProvider.Microsoft/CredentialProvider.Microsoft.csproj
+++ b/CredentialProvider.Microsoft/CredentialProvider.Microsoft.csproj
@@ -33,7 +33,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.IdentityModel.Clients.ActiveDirectory" Version="3.19.4" />
-    <PackageReference Include="NuGet.Protocol" Version="4.8.0-preview5.5326" />
+    <PackageReference Include="NuGet.Protocol" Version="5.2.0" />
     <PackageReference Include="PowerArgs" Version="3.0.0" />
   </ItemGroup>
 
--- a/CredentialProvider.Microsoft/CredentialProviders/VstsBuildTaskServiceEndpoint/VstsBuildTaskServiceEndpointCredentialProvider.cs
+++ b/CredentialProvider.Microsoft/CredentialProviders/VstsBuildTaskServiceEndpoint/VstsBuildTaskServiceEndpointCredentialProvider.cs
@@ -108,7 +108,7 @@ namespace NuGetCredentialProvider.CredentialProviders.VstsBuildTaskServiceEndpoi
             {
                 // Parse JSON from VSS_NUGET_EXTERNAL_FEED_ENDPOINTS
                 Verbose(Resources.ParsingJson);
-                Dictionary<string, EndpointCredentials> credsResult = new Dictionary<string, EndpointCredentials>();
+                Dictionary<string, EndpointCredentials> credsResult = new Dictionary<string, EndpointCredentials>(StringComparer.OrdinalIgnoreCase);
                 EndpointCredentialsContainer endpointCredentials = JsonConvert.DeserializeObject<EndpointCredentialsContainer>(feedEndPointsJson);
                 if (endpointCredentials == null)
                 {
--- a/CredentialProvider.Microsoft/Program.cs
+++ b/CredentialProvider.Microsoft/Program.cs
@@ -10,6 +10,7 @@ using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
+using NuGet.Common;
 using NuGet.Protocol.Plugins;
 using NuGetCredentialProvider.CredentialProviders;
 using NuGetCredentialProvider.CredentialProviders.Vsts;
@@ -142,10 +143,11 @@ namespace NuGetCredentialProvider
                             await RunNuGetPluginsAsync(plugin, multiLogger, TimeSpan.FromMinutes(2), tokenSource.Token).ConfigureAwait(continueOnCapturedContext: false);
                         }
                     }
-                    catch (TaskCanceledException)
+                    catch (OperationCanceledException ex)
                     {
                         // When restoring from multiple sources, one of the sources will throw an unhandled TaskCanceledException
-                        // if it has been restored successfully from a different source. We catch the exception and silently exit.
+                        // if it has been restored successfully from a different source.
+                        multiLogger.Log(LogLevel.Verbose, ex.ToString());
                     }
 
                     return 0;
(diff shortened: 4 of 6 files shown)


satbai:
The tests that are failing for you are succeeding on master branch on my Windows machine when running them with Visual Studio. Are you running the tests in a docker container? Can you share more details on how they're failing?

AMoghrabi:
Hey @satbai, I can definitely run the tests in Visual Studio again (also on Windows). I'll need to do this tomorrow if you don't mind; today is a holiday in Canada.

Thanks 😄

AMoghrabi:
Good morning @satbai

I am on the `master` branch with the latest pulled. Here are the two failed test results:

```
CanProvideCredentials_ReturnsTrueForKnownSources()

 Message: 
    Expected boolean to be true because http://example.pkgs.vsts.me/_packaging/TestFeed/nuget/v3/index.json is a known host, but found False.
  Stack Trace: 
    at LateBoundTestFramework.Throw(String message)
    at TestFrameworkProvider.Throw(String message)
    at DefaultAssertionStrategy.HandleFailure(String message)
    at AssertionScope.FailWith(String message, Object[] args)
    at BooleanAssertions.BeTrue(String because, Object[] becauseArgs)
    at <CanProvideCredentials_ReturnsTrueForKnownSources>d__10.MoveNext() in VstsCredentialProviderTests.cs line: 80
    at --- End of stack trace from previous location where exception was thrown ---
    at TaskAwaiter.ThrowForNonSuccess(Task task)
    at TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    at ThreadOperations.ExecuteWithAbortSafety(Action action)
```

```
CanProvideCredentials_ReturnsTrueForOverridenSources()

**Comment by satbai** [comment]
Could it be possible that your test runs detect any of [these three env vars](https://github.com/microsoft/artifacts-credprovider/blob/722f9480e1da88dd0a891e17d27d03a34f56373f/CredentialProvider.Microsoft/Util/EnvUtil.cs#L29)? 

When it comes to the timeout warnings, my guess is that since we're now ignoring the cancellation, the request that was going to be cancelled will hang until it times out. That's a pretty long timeout though, I don't know if we want to let the request hang for 100000ms. @zarenner I'd be interested in hearing your thoughts. 

Also, it would be worth a try to update your dotnet version to the latest and see if you still see the cancellation exceptions. NuGet 5.2 had some fixes in this area and according to the [release notes](https://docs.microsoft.com/en-us/nuget/release-notes/nuget-5.2-rtm), the [.NET SDK 2.2.40X](https://dotnet.microsoft.com/download/dotnet-core/2.2) will have the fixes.

**Comment by AMoghrabi** [comment]
That did the trick @satbai! I had `VSS_NUGET_EXTERNAL_FEED_ENDPOINTS` set on my machine. 

I'll get back to you on updating dotnet/NuGet later today or tomorrow. Appreciate the help, thanks!

**Comment by AMoghrabi** [comment]
Sorry for the delay on this. I upgraded my docker image to use dotnet SDK 2.2 and unfortunately I'm still seeing timeouts. 

Should I be installing NuGet 5.2 (assuming it's net core compatible) in my image and trying to restore with that?

**Comment by satbai** [comment]
I've seen people switching to the latest NuGet and the issue being resolved so if you have the opportunity to test NuGet restore, it would be interesting to see whether it'll fix it.

**Comment by zarenner** [comment]
@satbai @AMoghrabi
There was an issue on the NuGet side that is fixed with 5.2, but it's almost certainly unrelated to this issue (#108) as the callstack in this issue clearly shows the TaskCanceledException in the cred provider itself.

I don't have a good understanding of why exiting cleanly rather than crashing would introduce a timeout on the nuget side (or if that's related to this change). I'm starting to dig into understanding the plugin protocol better, but perhaps @dtivel has thoughts. @AMoghrabi https://github.com/NuGet/Home/wiki/Plugin-Diagnostic-Logging would be useful to gather as well, although until #121 goes in you'll only get the logs from the nuget side.

If we think that the timeouts are an unrelated issue, I'd be in favor of testing this PR a bit more and then merging it. I'm happy to help with that.

**Comment by AMoghrabi** [comment]
Thanks @zarenner. Im away until Monday so I'll check this out then. I would appreciate any help with testing. I would like this problem fixed soon so I can finish.moving my company to Azure artifacts.

Have a good weekend.

**Comment by zarenner** [comment]
@AMoghrabi I did a bit of testing yesterday and I didn't get any timeouts. However, I was tweaking some variables to force the TaskCanceledException issue to repro, so it wasn't a true test.

I'll be doing some more testing today - we have some customers here at Microsoft having similar looking failures I hope to test against. If it looks good I'll take your change, likely also with some additional logging in case this doesn't completely resolve the issue.

**Comment by AMoghrabi** [comment]
Ok great, thanks! I also want to circle back to Satbai's request to try the latest nuget.exe. Im not sure if I am unaware of something here, but my problem only occurs in a docker container, and I've been reproducing the error using the Dotnet tool. To my knowledge I cannot use nuget.exe in a Linux container unless I install mono which is not what we want.

However, if it is possible to use nuget.exe in a Linux container, I'll research and attempt to test that on Monday.

**Comment by AMoghrabi** [comment]
Hey @zarenner , sorry for the delay. I'll be hopping back onto this on Tuesday and will have an answer for you then.

**Comment by zarenner** [comment]
Sounds good, no worries. If the fix in the credprovider is sufficient for you, that's great - but there's definitely an issue on the NuGet side that may or may not make this change unnecessary (although I'm guessing it'll still be needed).

**Comment by AMoghrabi** [comment]
Hey @zarenner, I hope your weekend went well! 😄 

I've discovered a couple of things that I want to share. 

First, it looks like the image `microsoft/dotnet:2.1-sdk` hasn't been updated to use the latest NuGet version. I've changed that line in my docker to `mcr.microsoft.com/dotnet/core/sdk:2.1` and now it my docker build is using nuget version `5.2.0.3`. 

Unfortunately that does not change anything. The "official" published credential provider still throws the exception `Unhandled Exception: System.Threading.Tasks.TaskCanceledException: A task was canceled.` When I copy in the modified cred provider that includes my change, the error doesn't show up. 

Regarding the nuget timeouts, it appears that is a bug coming from NuGet itself, as you suspected as well. Since we don't have upstream enabled for Azure Artifacts, I decided to run the following command where `$NUGET_SOURCE` is our private server, which has upstream enabled:

```
RUN for f in ./*.csproj; do dotnet restore $f --source $NUGET_SOURCE --source https://api.nuget.org/v3/index.json; done
```

This throws timeouts. Example:

```
  Failed to download package 'Microsoft.TestPlatform.TestHost.16.0.1' from 'https://api.nuget.org/v3-flatcontainer/microsoft.testplatform.testhost/16.0.1/microsoft.testplatform.testhost.16.0.1.nupkg'.
  The download of 'https://api.nuget.org/v3-flatcontainer/microsoft.testplatform.testhost/16.0.1/microsoft.testplatform.testhost.16.0.1.nupkg' timed out because no data was received for 60000ms.
```

When I change the command to this (so there is only one source): 

```
RUN for f in ./*.csproj; do dotnet restore $f --source $NUGET_SOURCE
```

This works! No timeouts at all.

The timeouts don't cause the build to fail so it isn't a deal breaker. The build is slower to complete (obviously 😄). 

My only concern is this change may hide unexpected problems that throw a `TaskCanceledException`, and they shouldn't be handled by exiting silently. What are your thoughts on that?

zarenner:
@dtivel any idea if these timeouts look related (to the nuget/dotnet fixes we're testing)? Also, thoughts on still checking in this PR? I assume we'll still want to handle `TaskCanceledException` on the credprovider side even with any changes in nuget, but wanted to check.

AMoghrabi:
Hey @dvitel @zarenner , wondering if there are any updates on this. Thank you 😄

zarenner:
+jmyersmsft @nkolev92 while I (and I think also Damon) are on vacation

AMoghrabi:
Curious if anyone has had a chance to look at this yet. Thanks!

jmyersmsft:
A TaskCanceledException shouldn't be able to escape to your catch block -- if it does, there's a bug somewhere else. If a _request_ is canceled, the request handler infrastructure should swallow that and continue to be ready to handle requests. If the _connection_ is aborted for some reason, that may cause a lot of things to be canceled, but that still shouldn't bubble out (another exception might).

I'm going to prepare some bits for you that attempt to give some diagnostics about what got canceled and where by rummaging around in the guts of CancellationToken[Source]. If you can repro with those bits, that may help us get to the real source

jmyersmsft:
@AMoghrabi My branch [users/jmyers/cancelTracking](https://github.com/microsoft/artifacts-credprovider/tree/users/jmyers/cancelTracking) mixes your change, some further changes from @zarenner and my cancellation-tracking changes. I haven't tested the netfx version (should work, but my cancellation diagnostics code reflects into private implementation details), but the netcore version should work. If you could try to repro your failure with that code it may help us understand the problem. 

The cancellation diagnostics won't be displayed on the console, but if you set up a file log by setting the NUGET_CREDENTIALPROVIDER_LOG_PATH environment variable to an absolute path, the diagnostics will be written there.

If you need a real-signed build, I can get you one.

AMoghrabi:
Thanks Jonathan. I'll collect logs and send your way. Can I get your e-mail address?

jmyersmsft:
@AMoghrabi jmyers at microsoft dot com

jmyersmsft:
@AMoghrabi If you change the PR so it catches the base exception `OperationCanceledException` and logs it to `multilogger`, I will merge it, then layer some of the changes from my branch on top (some logging enhancements and some build-related stuff that has shifted underneath us), and make an official release

AMoghrabi:
Thanks Jon. I'll try getting that out today.

AMoghrabi:
My changes are in @jmyersmsft. This is obvious but I'll point it out anyways: I'm not sure if you're planning to add in your changes to `MultiLogger` so the `Log` method takes in a `allowOnConsole` argument. I've excluded that since those changes aren't in this branch.

Thanks!

PR #125, opened 2019-09-04
Improving README.md

satbai:
- Fix NuGet plugin discovery links
- Add sample dockerfile
- Add section that explains where it's shipped

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Install [Visual Studio version 15.9-preview1 or later](https://visualstudio.micr
 
 ## Setup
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -102,6 +107,9 @@ If you're running the command as part of an automated build on an unattended bui
 
 You may need to restart the agent service or the computer before the environment variables are available to the agent.
 
+### Docker containers
+You may use the example Dockerfile provided here: https://gist.github.com/shubham90/ad85f2546a72caa20d57bce03ec3890f
```
satbai: @shubham90 are you happy having this link here or should we create and example file in the repo?

satbai: Probably better to have an example file in the repo - do you think the helpers folder is sufficient or maybe a samples folder?

satbai: I'll make it a samples folder. Maybe there are other samples that would be helpful.

satbai: done.

PR #128, opened 2019-09-10
Clean up TLS1.2 check - Only download ZIP when necessary

BasilAtMicrosoft:
Simplify Net.ServicePointManager TLS1.2 check

Wrap download logic in a conditional - currently it downloads the zip even if it exists locally and -force isn't specified


jmyersmsft:
The ServicePointManager change is good, but I don't think the other part is actually doing anything -- Any time the conditional would evaluate to false, the script would already have been aborted by the `# Check if plugin already exists if -Force swich is not set` section directly above it

jmyersmsft:
We've decided not to take this PR at this time.

PR #139, opened 2019-10-15
Update README.md

sturlath:
An alternative way to "Windows+R"- run to get to the correct place of the nuget plugins.


satbai:
Thank you for your contribution!

Issue #143, opened 2019-10-30
Change exit timeout to start ticking on a shutdown request, not at startup


nkolev92:
The plugin lifetime in NuGet is managed by idleness. 
That means after a certain period of time, NuGet will shutdown the plugin process. 

When running in plugin mode, the plugin currently has a 2 min timeout, regardless of what happens. 
See: 
https://github.com/microsoft/artifacts-credprovider/blob/bea55735a710ede7460eee11f22f5700eac9c3f8/CredentialProvider.Microsoft/Program.cs#L225-L230

Large solution restores can sometimes last more than 2 minutes. In those cases, the plugin dies and reports an error to NuGet. 
NuGet logs that error to stderr. 

If NuGet needs the plugin for any other purpose throughout the lifetime of the operation, it will start another plugin process and do that successfully. 
So the overall operation succeeds, as far as NuGet is concerned.

It's a common for tooling/CIs to rely on the stderr as a signal of any issues, see: 
https://dev.azure.com/dnceng/internal/_build/results?buildId=405076
Windows Build UAP_x86_Release
That fails their whole build operation. 

I think the action from the plugin side should be the following:

1. When in plugin mode, never shutdown the process due to idleness. Allow the nuget counterpart to do that. 
The plugin infrastructure provides a way to monitor the process exit. 
1. Take special care when logging errors that are unrelated to current requests. 

//cc @jmyersmsft @dtivel @rrelyea

Note that all of the tests I have done contain 4 NuGet side fixes. If you were to take stock nuget(5.3 and older) + plugin you might not experience this. 

Related fixes: 

1.	NuGet internally cancelling prematurely. (5.3.1) https://github.com/NuGet/Home/issues/8648
2.	Logging failures triggering plugin shut down. (5.3.1) https://github.com/NuGet/Home/issues/8688
3.	Propagated exception at dispose time fails the whole auth with a task cancellation error. (5.4) https://github.com/NuGet/Home/issues/8732
4.	Plugin logging accuracy improvement (5.4) https://github.com/NuGet/Home/issues/8747

Related issue: https://github.com/NuGet/Home/issues/8528

jmyersmsft:
I don't think I wrote the offending line originally, so I don't know the original intent, but I think the right behavior is to make it a _shutdown_ timeout, not an overall operation timeout

nkolev92:
> make it a shutdown timeout,

Can you please elaborate?

jmyersmsft:
[...]

PR #145, opened 2019-11-01
Don't start ticking the shutdown timer until BeforeClose

jmyersmsft:


jmyersmsft:
Fixes #143

CredentialProvider.Microsoft/Program.cs
```diff
@@ -208,23 +208,38 @@ public static async Task<int> Main(string[] args)
             }
         }
 
-        internal static async Task RunNuGetPluginsAsync(IPlugin plugin, ILogger logger, TimeSpan timeout, CancellationToken cancellationToken)
+        internal static async Task WaitForPluginExitAsync(IPlugin plugin, ILogger logger, TimeSpan shutdownTimeout)
```
satbai: Would it be useful to have the opportunity to set the timeout as an environment variable and have 2 minutes as a default?

PR #157, opened 2019-11-28
Add version support to .sh script

satbai:


satbai:
Works with Windows - pending testing with Linux/Mac

aasim:
This is no longer relevant as onprem has since been updated to automatically have the cred provider

PR #183, opened 2020-04-15
Add override for CanShowDialog

jmyersmsft:


CredentialProvider.Microsoft/Util/EnvUtil.cs
```diff
@@ -75,6 +76,17 @@ public static IList<string> GetHostsFromEnvironment(ILogger logger, string envVa
             return hosts;
         }
 
+        public static bool? ForceCanShowDialogTo()
+        {
+            var fromEnv = Environment.GetEnvironmentVariable(ForceCanShowDialogEnvVar);
+            if(string.IsNullOrWhiteSpace(fromEnv) || !bool.TryParse(fromEnv, out var parsed))
+            {
+                return default;
```
satbai, line 84: Just checked the default value of a nullable bool. I think it's null which is what we want here but sometimes I can't understand English. 

"An instance for which the HasValue property is false and the Value property is undefined. That default value is also known as the null value of a nullable value type."
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/default-values

CredentialProvider.Microsoft/Util/EnvUtil.cs
```diff
@@ -29,6 +29,7 @@ public static class EnvUtil
         public const string BuildTaskUriPrefixes = "VSS_NUGET_URI_PREFIXES";
         public const string BuildTaskAccessToken = "VSS_NUGET_ACCESSTOKEN";
         public const string BuildTaskExternalEndpoints = "VSS_NUGET_EXTERNAL_FEED_ENDPOINTS";
+        public const string ForceCanShowDialogEnvVar = "ARTIFACTS_CREDPROVIDER_FORCE_CANSHOWDIALOG_TO";
```
satbai, line 32: Add this to the EnvironmentVariableHelp help string that has all the other env vars.

CredentialProvider.Microsoft/Util/EnvUtil.cs
jmyersmsft: Yes. `default` where an expression of type `Nullable<bool>` is expected is equivalent to `null` in the same place, which are both syntactic sugar for `default(Nullable<bool>)`. The reason for the strange wording in the doc is that `Nullable<T>` is still a value type, so it technically can't be null per se, but syntactic sugar in the language sort of hides that.

CredentialProvider.Microsoft/Util/EnvUtil.cs
satbai: and to readme manually :(

jmyersmsft: I low-key hate those lists -- they're full of crappy implementation details, bad legacy decisions, and weird inconsistencies, especially since this is used for more than just nuget now

satbai: Also maybe add a note there that using this env var is not normally recommended.

satbai: This is somewhat unrelated to this PR but maybe it would make sense to only have the environment variables in the list that we expect people to actually use. Or at least separate the nuget specific variables form the "universal" ones. We could have the full list somewhere else but keep the help string clean.

jmyersmsft: I don't know if I'll ever have time, but I'd like to split the whole thing up into CredentialProvider.Microsoft.exe (NuGet plugin + back-compat wrapper) and artifacts-credprovider.exe (all other use-cases, with all the variables and options renamed and rearranged to actually make sense)

satbai: That would make a lot of sense

PR #185, opened 2020-04-21
Add msal support

crmann1:
Add MSAL support to use a newer library from the AAD auth team which will give access to the newest auth stacks available for nuget.exe and others using the provider.

Has two supported UI flows:
1) Device code flow
2) System browser flow
This flow allows the use of the system web browser as an interactive flow. This was tested on OSX, Linux and Windows. 

Fall back for token acquisition:
1) Cache on disk if present
2) MSAL windows integrated authentication
3) System browser flow if can show dialog is true
4) MSAL device code flow

The MSAL cache file if enabled only persists on windows and uses DPAPI for encryption. 

By default ADAL is still used. But to enable MSAL do the following
Set environment variable  
NUGET_CREDENTIALPROVIDER_MSAL_ENABLED=true

To enable a persistent MSAL cache in the same location as the session token cache .
NUGET_CREDENTIALPROVIDER_MSAL_FILECACHE_ENABLED=true


CredentialProvider.Microsoft/CredentialProviders/Vsts/IAuthUtil.cs
```diff
@@ -70,7 +72,9 @@ public async Task<Uri> GetAadAuthorityUriAsync(Uri uri, CancellationToken cancel
             // Return the common tenant
             var aadBase = UsePpeAadUrl(uri) ? "https://login.windows-ppe.net" : "https://login.microsoftonline.com";
             logger.Verbose(string.Format(Resources.AADAuthorityNotFound, aadBase));
-            return new Uri($"{aadBase}/common");
+
+            var tenant = EnvUtil.MsalEnabled() ? OrganizationsTenant: CommonTenant;
```
jmyersmsft, line 76: Does MSAL not support MSA users? Azure DevOps is not limited to AAD customers. In fact, given the code just above this, it's likely only MSA-backed orgs get to this part,

CredentialProvider.Microsoft/Resources.resx
```diff
@@ -355,7 +355,27 @@ Device Flow Authentication Timeout
 NuGet workarounds
     {15}
         Set to "true" or "false" to override any other sources of the 
-        CanShowDialog parameter</value>
+        CanShowDialog parameter
+
+Enable MSAL
```
jmyersmsft, line 360: Could you put "(preview)" after the title?

Build.props
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <CredentialProviderVersion>0.1.22</CredentialProviderVersion>
+    <CredentialProviderVersion>0.1.23</CredentialProviderVersion>
```
jmyersmsft, line 4: Be sure to update this after merging/rebasing away the conflicts

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalBearerTokenProviders.cs
```diff
@@ -0,0 +1,128 @@
...
+
+                        return Task.CompletedTask;
+                    },
+                    cancellationToken,
+                    this.msalTokenProvider.Logger))?.AccessToken;
+        }
+
+        public bool ShouldRun(bool isRetry, bool isNonInteractive, bool canShowDialog)
+        {
+            return !isNonInteractive;
```
nayanshah, line 125: Assuming #203 gets merged first, please update the MSAL implementation as well.

jmyersmsft:
@herenhuang I have already provided feedback on this PR, which has not been addressed.

CredentialProvider.Microsoft/CredentialProviders/Vsts/IAuthUtil.cs
crmann1: Since you are using the visual studio appid which uses MSA passthrough, you cannot sign into AAD on the V2 endpoint using common. You must use organizations. 

If You were to get a different clientID which does not use MSA passthrough then you can sign in using consumers or common.

jmyersmsft: If we got a new app id without "MSA passthrough", how would that affect MSA users? Would they still be able to sign in to Azure DevOps through this tool? How about users who still have both MSA and AAD identities with the same UPN?

herenhuang:
Hey @crmann1 - gentle reminder we added some comments! :)

PR #188, opened 2020-05-11
Added logging before ADAL UI flow.

apluche:
Adding some logging for the ADAL UI so the logging catches that there may need to be a manual sign-in.

### DIFF
--- a/CredentialProvider.Microsoft/CredentialProviders/Vsts/BearerTokenProviders.cs
+++ b/CredentialProvider.Microsoft/CredentialProviders/Vsts/BearerTokenProviders.cs
@@ -83,7 +83,7 @@ namespace NuGetCredentialProvider.CredentialProviders.Vsts
 
         public async Task<string> GetTokenAsync(Uri uri, CancellationToken cancellationToken)
         {
-            logger.Minimal(string.Format(Resources.UIFlowStarted, this.Name));
+            logger.Minimal(string.Format(Resources.UIFlowStarted, this.Name, uri.AbsoluteUri));
             return (await adalTokenProvider.AcquireTokenWithUI(cancellationToken))?.AccessToken;
         }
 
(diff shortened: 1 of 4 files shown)


CredentialProvider.Microsoft/Resources.resx
```diff
@@ -456,4 +456,7 @@ NuGet workarounds
   <data name="ForcingCanShowDialogFromTo" xml:space="preserve">
     <value>Forcing canShowDialog from {0} to {1} due to setting from environment variable</value>
   </data>
+  <data name="UIFlowStarted" xml:space="preserve">
+    <value>Using the {0}  flow. User sign-in required in pop-up authentication window.</value>
```
satbai, line 460: nit: User sign-in required in _a_ pop-up authentication window.

CredentialProvider.Microsoft/CredentialProviders/Vsts/BearerTokenProviders.cs
```diff
@@ -70,17 +70,20 @@ public bool ShouldRun(bool isRetry, bool isNonInteractive, bool canShowDialog)
...
+            this.adalTokenProvider = adalTokenProvider;
+            this.logger = logger;
         }
 
         public bool Interactive { get; } = true;
         public string Name { get; } = "ADAL UI";
 
         public async Task<string> GetTokenAsync(Uri uri, CancellationToken cancellationToken)
         {
+            logger.Minimal(string.Format(Resources.UIFlowStarted, this.Name));
```
satbai, line 86: Would it be useful to have the URI in the log too? Nah probably not necessary

apluche: I think the main point is to know that a pop-up is occurring rather than know the specific instance of it. I can add that in the logging just so we won't have to do it later.

CredentialProvider.Microsoft/Resources.resx
apluche: yikes, nice catch. Fixing it now

Issue #136, opened 2019-10-10
Where is proxy parameters?


VerdonTrigance:
Guys, seriously! All other companies do this stuff for their own distribution scripts, but Microsoft - never! It's not funny! How to install this cred provider under the proxy that require authentication?!
I mean this script: https://github.com/microsoft/artifacts-credprovider/blob/master/helpers/installcredprovider.ps1
PS: I also tried to use System.Net.WebClient but faced with problem at downloading $releaseUrl
`Unable to resolve the Credential Provider zip file from https://api.github.com/repos/Microsoft/artifacts-credprovider/releases/latest`

tkrick:
filed an internal work item for us to add this capability. we'll take a look.

VerdonTrigance:
Any updates?

satbai:
I have a PR out to add proxy support. I just need to test it behind a proxy to make sure it works as expected.  If you're behind a proxy feel free to check the PR or try the script [here](https://github.com/microsoft/artifacts-credprovider/blob/4ce28570bfc06152dfad23c948a47a02635c73ae/helpers/installcredprovider.ps1).

VerdonTrigance:
> I have a PR out to add proxy support. I just need to test it behind a proxy to make sure it works as expected. If you're behind a proxy feel free to check the PR or try the script [here](https://github.com/microsoft/artifacts-credprovider/blob/4ce28570bfc06152dfad23c948a47a02635c73ae/helpers/installcredprovider.ps1).

Thank you for PR, but I am trying to Chocolatey whenever is possible and avoid Microsoft package and artifact managers.

satbai:
Thanks! Just for anyone looking for answers here, the correct place to set proxy parameters for .NET is the system proxy config.

PR #199, opened 2020-07-01
Adding proxy support

satbai:
Addressing issue https://github.com/microsoft/artifacts-credprovider/issues/136


jmyersmsft:
The correct place to set proxy parameters for .NET is the system proxy config. We should not add additional confusing and likely incomplete proxy configuration parameters to our ,NET tools except where interactions with external code requires it.

PR #203, opened 2020-07-30
Allow disabling DeviceFlow authentication using an environment variable

nayanshah:
Implementation is similar to one used for turning off Windows Integrated Auth.

### DIFF
--- a/CredentialProvider.Microsoft/CredentialProviders/Vsts/BearerTokenProviders.cs
+++ b/CredentialProvider.Microsoft/CredentialProviders/Vsts/BearerTokenProviders.cs
@@ -133,7 +133,7 @@ namespace NuGetCredentialProvider.CredentialProviders.Vsts
 
         public bool ShouldRun(bool isRetry, bool isNonInteractive, bool canShowDialog)
         {
-            return !isNonInteractive;
+            return !isNonInteractive && EnvUtil.DeviceFlowAuthenticationEnabled();
         }
     }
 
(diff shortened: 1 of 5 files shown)


nayanshah:
@johnterickson , @jmyersmsft , @satbai (mentioning since I can't assign reviews)

jmyersmsft:
I'm not sure that this is the best solution to the problem. If you need to avoid interactive methods, there's already a switch for that. Device flow is the very last thing we try, so if it's skipped, the overall credential request will just fail.

nayanshah:
Device flow errors in build logs causes confusion since people attempt to use the code from the logs even though request has already timed out. My plan is to turn Device flow off by default and rely on Windows Integrated for silent & ADAL UI for interactive auth. 

P.S: I understand this scenario mainly works on AD joined Windows boxes and might have to re-consider defaults in future.

nayanshah:
@jmyersmsft , @satbai , @apluche : any other feedback?

satbai:
Like Jonathan mentioned earlier, device flow is the last option which means by the time we get to device flow the other methods already didn't work out for some reason. 

The message in the case of device flow timeout is "Device flow authentication failed. User was presented with device flow, but didn't react within {0} seconds.". Perhaps rather than disabling device flow we could make logging better and add "Please try again and react to device flow within the provided timeout {0} to authenticate." To make it more clear that device flow only works within the given timeout period. Additionally, we could mention the timeout period in the initial device flow message as well. The timeout value can be modified with an environment variable.

jmyersmsft:
We have decided not to take this change at this time, sorry.

PR #211, opened 2020-09-09
fix(credentials): fix credentials for feeds with space in url

### LINKED ISSUE #210: Plugin cannot find credentials in case feed URL contains spaces
We are running on-prem DevOps Server 2019 SP1 and b/c our collection name contains space the NuGet feed URL looks like:
https://site.com/tfs/Dev Collection/_packaging/our-feed/nuget/v3/index.json

in our build pipeline, we have added NuGetAuthenticate task that sets required environment variables:
##[debug]set VSS_NUGET_EXTERNAL_FEED_ENDPOINTS=
{"endpointCredentials":
  [
    {"endpoint":"https://site.com/tfs/Dev Collection/_packaging/our-feed/nuget/v3/index.json","password":***}
  ]
}

But the next task that is running `dotnet restore` if failing with error:
[CredentialProvider]VstsBuildTaskServiceEndpointCredentialProvider - Environment variable VSS_NUGET_EXTERNAL_FEED_ENDPOINTS did not have credentials for endpoint https://site.com/tfs/Dev%20 Collection/_packaging/our-feed/nuget/v3/index.json (TaskId:57)

After digging the sources of credentials provider I can see it is trying to much the URL exactly, but it takes the URL to much from 
`string uriString = request.Uri.AbsoluteUri;`

which returns URL-encoded value string (where space is replaced with %20) and as a result it can't find the value in environment variable b/c URLs doesn't much exactly:
`bool endpointFound = Credentials.TryGetValue(uriString, out EndpointCredentials matchingEndpoint);`

So,  probably the best way to solve this would be to fix ParseJsonToDictionary() method and put URL-encoded value into Credentials dictionary.

luber:
fix issue with with missing credentials in case feed url contained spaces

fix #210

### DIFF
--- a/Build.props
+++ b/Build.props
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <CredentialProviderVersion>0.1.24</CredentialProviderVersion>
+    <CredentialProviderVersion>0.1.25</CredentialProviderVersion>
   </PropertyGroup>
 </Project>
--- a/Build.props
+++ b/Build.props
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <CredentialProviderVersion>0.1.25</CredentialProviderVersion>
+    <CredentialProviderVersion>0.1.26</CredentialProviderVersion>
   </PropertyGroup>
 </Project>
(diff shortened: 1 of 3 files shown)


luber:
@satbai could you please review this?

luber:
@jmyersmsft could you please review this?

Build.props
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <CredentialProviderVersion>0.1.24</CredentialProviderVersion>
+    <CredentialProviderVersion>0.1.25</CredentialProviderVersion>
```
jmyersmsft, line 4: 0.1.25 has already been released. Please make this 0.1.26

jmyersmsft (approved):

PR #222, opened 2020-11-20
Update README with additional mono msbuild notes

### LINKED ISSUE #221: installcredprovider unix script doesn't install netfx plugin
The README says to use the installcredprovider.sh script to get the latest cred provider. However, this script only installs the netcore plugin, meaning the credential provider won't work with mono msbuild, which requires the netfx version.

Also the READMe states that one can use wget. However, wget is not present on macOS per default and has to be installed before trying this. Perhaps updating the command to curl would help as (I think) it is more common on unix systems.

```sh
/bin/bash -c "$(curl -fsSL https://aka.ms/install-artifacts-credprovider.sh)"
```

If you want, I can make a proposal on an update to the README to provide these details for anyone encountering the same issue.

Cheesebaron:
Updated README with some notes about what to do for mono msbuild, as the installcredprovider.sh script only installs the netcore plugin on unix based environments.

Also updated the example on how to download and run. `wget` is not present on macOS per default, so showing an example using `curl` instead.

Fixes #221 partially. You might consider changing the script to _also_ install the netfx plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,10 @@ Using the above is recommended, but as per [NuGet's plugin discovery rules](http
 #### Automatic bash script
 
 [Linux or Mac helper script](helpers/installcredprovider.sh)
-- e.g. `sh -c "$(curl -fsSL https://aka.ms/install-artifacts-credprovider.sh)"`
+
+Examples:
+- `wget -qO- https://aka.ms/install-artifacts-credprovider.sh | bash`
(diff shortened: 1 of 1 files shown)


shadargee1982:
@Cheesebaron - I think keeping the wget option doesn't hurt.  I suggest that you update the docs so both command options are present.

Cheesebaron:
Better?

PR #268, opened 2021-09-13
Updating version to .NET Core 3.1

alexdegroot:
Since clean developer machines no longer contain .NET Core 2. I've upgraded all the bits to .NET Core 3.
Please let me know what I can do to get this change in.

### DIFF
--- a/Build.props
+++ b/Build.props
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <CredentialProviderVersion>0.1.27</CredentialProviderVersion>
+    <CredentialProviderVersion>0.2.0</CredentialProviderVersion>
   </PropertyGroup>
 </Project>
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ In this and subsequent examples, configuration can be either `debug` or `release
 ### Publishing
 
 ```shell
-dotnet publish CredentialProvider.Microsoft --configuration Release --framework netcoreapp2.1
+dotnet publish CredentialProvider.Microsoft --configuration Release --framework netcoreapp3.1
 ```
 
 ### Packing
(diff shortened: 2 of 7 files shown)


satbai:
Thanks for the PR. We are going to make this change because the .NET Core version we're using is now out of support. However, we need to do some testing and investigation because this will be a breaking change. This work has been added to the team backlog and will be going through triage. Thanks again!

alexdegroot:
Let me in the meanwhile resolve the dependencies and add what testing I've done myself

johnterickson:
This is certainly the right direction, but we need to be a bit careful here because we have instructed folks to run `wget -qO- https://aka.ms/install-artifacts-credprovider.sh | bash` which will automatically pull the latest.  If they have `dotnet` 2 installed, then this is working fine.  But, if we bump our build to build against `dotnet` 3 and they don't have it installed, then those (CI?) builds will start failing all of a sudden.

sandyarmstrong:
Any movement here? This has been a big problem for our team.

sandyarmstrong:
> This is certainly the right direction, but we need to be a bit careful here because we have instructed folks to run `wget -qO- https://aka.ms/install-artifacts-credprovider.sh | bash` which will automatically pull the latest. If they have `dotnet` 2 installed, then this is working fine. But, if we bump our build to build against `dotnet` 3 and they don't have it installed, then those (CI?) builds will start failing all of a sudden.

@johnterickson yes, builds will fail, but it should be easy to add a check in https://github.com/microsoft/artifacts-credprovider/blob/master/helpers/installcredprovider.sh to make sure the SDK is available in `/usr/local/share/dotnet` (or `$DOTNET_ROOT` on Azure DevOps). Then the build output can give easy-to-follow guidance to fix the situation.

You'll also want to bump to .NET 6, since that is the only fully supported SDK version on M1 right now, and is your best bet for proper support for users on newer Macs.

sandyarmstrong:
@satbai I see you are working on this in https://github.com/microsoft/artifacts-credprovider/commits/users/satbai/netcoreversion. Please increase the TargetFramework from .NET Core 3.1 to .NET 6 to better support M1 users. Thanks!

PR #269, opened 2021-09-13
Use newtonsoft 13.0.1 and remove myget feed as a package source

satbai:
- Use 13.0.1 version of Newtonsoft
- remove myget feed from nuget.config
- bump version

### DIFF
--- a/Build.props
+++ b/Build.props
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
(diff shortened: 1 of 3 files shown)


apluche (approved):

satbai:
The failing release build is me manually running a test build on my branch since we're planning on releasing a new version (the last release was November 2020 so a long time ago!). I'm going to merge the PR and see if it fails on the master branch as running it on a dev branch is not the normal workflow. To run a successful release build on master, the change in nuget.config in this PR must be included, otherwise it fails. If the build fails on master, I will revert this and merge the nuget.config change separately to investigate the failure more.

PR #277, opened 2021-11-09
Update installcredprovider.ps1 to support Powershell Core in NanoServer container

bladedeyna:
Adds support for Powershell Core when running in a Docker NanoServer container. 

The GetFolderPath function does not resolve correctly in this environment, however `$env:UserProfile` does. So this PR adds a fallback if the `[System.Environment]::GetFolderPath([System.Environment+SpecialFolder]::UserProfile` function returns an empty string, to use the `$env:UserProfile` environment variable instead.

image

### DIFF
--- a/helpers/installcredprovider.ps1
+++ b/helpers/installcredprovider.ps1
@@ -22,7 +22,12 @@ if ([Net.ServicePointManager]::SecurityProtocol.ToString().Split(',').Trim() -no
     [Net.ServicePointManager]::SecurityProtocol += [Net.SecurityProtocolType]::Tls12
 }
 
-$profilePath = [System.Environment]::GetFolderPath([System.Environment+SpecialFolder]::UserProfile)
+if ([System.Environment]::GetFolderPath([System.Environment+SpecialFolder]::UserProfile) -ne '') {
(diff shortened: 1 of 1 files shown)


helpers/installcredprovider.ps1
```diff
@@ -22,7 +22,12 @@ if ([Net.ServicePointManager]::SecurityProtocol.ToString().Split(',').Trim() -no
     [Net.ServicePointManager]::SecurityProtocol += [Net.SecurityProtocolType]::Tls12
 }
 
-$profilePath = [System.Environment]::GetFolderPath([System.Environment+SpecialFolder]::UserProfile)
+if ([System.Environment]::GetFolderPath([System.Environment+SpecialFolder]::UserProfile) -ne '') {
```
aasim, line 25: Can you assign the return value of GetFolderPath to a variable and then check for it being empty? That way you don't  have to call that function twice.

aasim (changes requested):

helpers/installcredprovider.ps1
bladedeyna: Thanks for your review, I've added the changes requested.

aasim (approved):

PR #281, opened 2021-11-29
Update stale.yml

saishiva0691:


edgarrs:
Why is this change needed?

PR #285, opened 2022-01-05
Update MicroBuild tasks in build

jmyersmsft:
Reaction to required move to new MicroBuild pools for real-signed builds


satbai (approved):
lgtm

PR #293, opened 2022-01-26
Vsts doesn't handle pkgs.dev.azure.com feeds hence cache is not working

DavidKarlas:
See https://github.com/dotnet/templating/issues/4278 for more info


DavidKarlas:
Ok, I'm not so sure about this anymore, closing for now...

PR #295, opened 2022-02-10
changed install from cli example

Carnes:
**Update**
This PR is a slight change to the instructions to make it clear that the powershell script example can only be run from powershell.exe.  I've been running powershell scripts and commands from cmd.exe for many years and this actually tripped me up.  Thank you for your time.

**Previous comment:**
Description:
Changed the cli install example to use wget to fetch and then run the script.  This command downloaded and ran the installer without issue.  Also included powershell in the example because not every dev would know to prefix that without being told to do so.  Typically a command presented as a code snippet should be complete and not expect the user to make changes to it (defeating the purpose of a copy/paste command).

Reason for change:
Received an "Access Denied" error when attempting to use the install example from cli.  Also tried it as an administrator.  Found a Microsoft article explaining that iex shouldn't be used: https://devblogs.microsoft.com/powershell/invoke-expression-considered-harmful/
image

Please let me know if there are any alterations needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Dotnet needs the `netcore` version to be installed. NuGet and MSBuild need the `
 
 [PowerShell helper script](helpers/installcredprovider.ps1)
 - To install netcore, run `installcredprovider.ps1`
-  - e.g. `iex "& { $(irm https://aka.ms/install-artifacts-credprovider.ps1) }"`
+  - e.g. `powershell wget https://aka.ms/install-artifacts-credprovider.ps1 -OutFile install-artifacts-credprovider.ps1 && powershell ./install-artifacts-credprovider.ps1`
 - To install both netfx and netcore, run `installcredprovider.ps1 -AddNetfx`. The netfx version is needed for nuget.exe.
   - e.g. `iex "& { $(irm https://aka.ms/install-artifacts-credprovider.ps1) } -AddNetfx"`
 
--- a/README.md
+++ b/README.md
@@ -37,10 +37,12 @@ Dotnet needs the `netcore` version to be installed. NuGet and MSBuild need the `
 #### Automatic PowerShell script
 
 [PowerShell helper script](helpers/installcredprovider.ps1)
+- Use powershell.exe to perform the following steps
 - To install netcore, run `installcredprovider.ps1`
-  - e.g. `powershell wget https://aka.ms/install-artifacts-credprovider.ps1 -OutFile install-artifacts-credprovider.ps1 && powershell ./install-artifacts-credprovider.ps1`
+  - e.g. `iex "& { $(irm https://aka.ms/install-artifacts-credprovider.ps1) }"`
 - To install both netfx and netcore, run `installcredprovider.ps1 -AddNetfx`. The netfx version is needed for nuget.exe.
   - e.g. `iex "& { $(irm https://aka.ms/install-artifacts-credprovider.ps1) } -AddNetfx"`
+
 
 #### Manual installation on Windows
 


OskarKlintrot:
Running powershell from cmd feels a bit odd when you can just open powershell instead of cmd..? Also, wget isn't available on pwsh.

Carnes:
A cli example is assumed to be run from a standard terminal (like cmd).  If the cli example cannot be run from cmd.exe then it really needs call out what specific program that command is for.

The point of that section of documentation is how to install the plugin.  It should be clear how to do it.  If it's preferred to stick with iex (which the MS blog says is harmful) then the text could be updated to say how to run it.  Either way is good, imo.  Just looking for clear instructions.

Carnes:
Updated to just add a clarifying step that powershell.exe should be used as the standard terminal (cmd.exe) won't work.

OskarKlintrot:
Powershell and cmd are shells, not terminals.

OskarKlintrot:
The subtitle says "Automatic PowerShell script" so I don't think you have to state again that it should be run in Powershell.

Carnes:
If running the commands within powershell are a requirement then that should be stated.  The linux installation instructions explicitly use bash, for example.  If you try that from cmd.exe by prefixing with poweshell then it doesn't work.

@OskarKlintrot the instructions are not clear.  I'm open to suggestions for what can be changed to clear it up.

OskarKlintrot:
So the instruction under the heading "Automatic PowerShell script" isn't clear that you should indeed use powershell to run the powershell script? The instructions for bash is identical, they just changed "powershell" to "bash".

Carnes:
Yes, you are exactly correct.  It is not clear.  Normally when you need to run powershell from cmd.exe you just prefix with the word powershell.  Example: "powershell wget https://aka.ms/install-artifacts-credprovider.ps1".  However, that does not work with the provided iex code.

I thought the provided script was broken when I first tried it and it didn't work.  That's why i suggested the change to wget.  It never occurred to me that you couldn't run it in cmd.exe and you had to use powershell.exe.  Not until you pointed that out.  I then changed the code back and suggested just adding that fact to the instructions.

The instructions for bash are not identical, like you suggest.  They prefix/pipe with bash which allows you to run it from any shell.  That is the issue with the powershell instructions.  You cannot just prefix with powershell and use it from the standard shell (cmd.exe).  Please let me know if you don't understand this.

OskarKlintrot:
I hear you but I still think you are wrong. If you tell someone to run a Powershell script and they try to do it any other way than open Powershell and run the script they are on their own. You can run it via cmd, bash, from `win+r`, the nuget package etc etc. Should all those scenarios be covered? Of course not. If you need to install the cred provider, I definitely think the team is right to assume that the user has the know-how to run a Powershell script if instructed to. And remember that `powershell.exe` is v5 and below, v6 and above is `pwsh`. So, the PR says that you must use Powershell v5, which is incorrect. Powershell 6-7 works fine as well.

Btw I just tried to use cmd and it works:
image

The issue seems to be something else for you. If I tried to pass the script as an arg to powershell I get a different error than you get.

image


Anyway, let's agree to disagree. Over and out.

Carnes:
@OskarKlintrot it looks like you are intentionally wasting the time of anyone trying to read this PR.  You are disagreeing with something I am not proposing.  You clearly are knowledgeable of powershell but have not used that to help improve this PR at all.  

The strangest part is that we're actually in agreement that the script can only be run from powershell.exe.  You even demonstrated that fact.  Let's add that fact to the instructions and make it clear.

OskarKlintrot:
> you are intentionally wasting the time of anyone trying to read this PR
1. You proposed a change that would not work on Powershell v6-7. I pointed that out and you reverted it.
2. I've been trying a few times now (not clearly enough though) to make you aware that `powershell.exe` is v5 and below and `pwsh.exe` is v6 and above. The docs work for all versions I've tested (v5 and above) and not only `powershell.exe`.

Carnes:
> You proposed a change that would not work on Powershell v6-7. I pointed that out and you reverted it.
This did not happen.  The only thing changed was removing the cmd.exe style of execution and just adding a clarifying instruction.  You can see that in the commit chain.  That was actually a good thing you pointed out.  You did not suggest an improvement so I took a guess at a fix based on your comment.

> I've been trying a few times now (not clearly enough though) to make you aware that powershell.exe is v5 and below and pwsh.exe is v6 and above. The docs work for all versions I've tested (v5 and above) and not only powershell.exe.
I'm understanding what you are saying.  You are not suggesting any change at all.  You are just criticizing.  Please, suggest an improvement based on your understanding of the issue.  

It sounds like you would prefer the clarifying instruction to be changed to "Use powershell.exe or pwsh.exe to perform the following steps".  Is that correct?  That looks like a positive change.

satbai:
Thanks for the PR. We're deciding not to take this change at this time.

PR #298, opened 2022-02-24
Allow installation of releases other than latest

reillysiemens:
The current PowerShell installation helper allows users to install a specific version of the credential provider, but the same capability is not extended to the Bash version.

This change allows users to set an environment variable to specify a version other than `latest`. I changed the name of the `$VERSION` environment variable to decrease the likelihood of a conflict with the user's environment.

#### Installing `latest`
```bash
curl -fsSL https://aka.ms/install-artifacts-credprovider.sh
```

#### Installing `0.1.25`
```bash
export AZURE_ARTIFACTS_CREDENTIAL_PROVIDER_VERSION='0.1.25'
curl -fsSL https://aka.ms/install-artifacts-credprovider.sh
```

### DIFF
--- a/helpers/installcredprovider.sh
+++ b/helpers/installcredprovider.sh
@@ -6,9 +6,9 @@
 
 REPO="Microsoft/artifacts-credprovider"
 FILE="Microsoft.NuGet.CredentialProvider.tar.gz"
-VERSION="latest"
+: ${AZURE_ARTIFACTS_CREDENTIAL_PROVIDER_VERSION="latest"}
 # URL pattern documented at https://help.github.com/en/articles/linking-to-releases as of 2019-03-29
-URI="https://github.com/$REPO/releases/$VERSION/download/$FILE"
+URI="https://github.com/$REPO/releases/${AZURE_ARTIFACTS_CREDENTIAL_PROVIDER_VERSION}/download/$FILE"
 NUGET_PLUGIN_DIR="$HOME/.nuget/plugins"
 
 # Ensure plugin directory exists


kyle-rader:
I'll note the motivation for this is due in part to `latest` (as of writing, `0.1.28`) having broken interactive web auth via MSAL leaving only Device Code Flow, which is less than ideal. So installing `0.1.25` is currently preferrable.

satbai:
Thanks for the PR! It's a great idea to add support for installing other versions than the latest. 

However, using an environment variable seems a little odd - is that a normal way to pass parameters to a script in linux/mac? Wouldn't using something like a `-version` flag be better? I'm not a daily user or linux/mac so this may just be my ignorance and if so, please let me know. Thanks!

helpers/installcredprovider.sh
```diff
@@ -6,9 +6,9 @@
 
 REPO="Microsoft/artifacts-credprovider"
```
satbai, line 7: Please also add the instructions on how run the script to get a specific version to the comment section above so that the script so users can see that it's supported.

kyle-rader:
@satbai Using environment variables as config override is a fairly standard way to design any CLI program or script, and not just for CLIs but also services. This is a core tenant of 12 factor app design: https://12factor.net/config. 

Using option parameters specifically in scripts on Linux/Mac is generally less common in my experience though not unheard of. (If we did want to do that, we would want to use a GNU standard option of double dash `--<option-name>`.)

+1 to adding use of this override in the documentation.

reillysiemens:
@satbai, thanks for the quick response!

As @kyle-rader mentioned, using environment variables for this kind of thing is fairly standard (at least in the Linux/macOS world). A `--version` flag would also do the trick and might be preferable under different circumstances.

In this case, however, the script is meant to be consumed via a "trust fall" download that's piped to a shell and passing flags/arguments to such scripts is usually more syntactically cumbersome for users.

I'll update the comment to reflect this as a supported feature and recommend going with an environment variable for simplicity now. It might be worthwhile at a later point to review the [PowerShell script](https://github.com/microsoft/artifacts-credprovider/blob/master/helpers/installcredprovider.ps1) and try to bring this script into parity with the options that one has.

helpers/installcredprovider.sh
```diff
@@ -3,12 +3,16 @@
...
+# variable to match the name of a supported release.
+# SEE: https://github.com/microsoft/artifacts-credprovider/releases
 
 REPO="Microsoft/artifacts-credprovider"
 FILE="Microsoft.NuGet.CredentialProvider.tar.gz"
-VERSION="latest"
+: ${AZURE_ARTIFACTS_CREDENTIAL_PROVIDER_VERSION="latest"}
 # URL pattern documented at https://help.github.com/en/articles/linking-to-releases as of 2019-03-29
-URI="https://github.com/$REPO/releases/$VERSION/download/$FILE"
+URI="https://github.com/$REPO/releases/${AZURE_ARTIFACTS_CREDENTIAL_PROVIDER_VERSION}/download/$FILE"
```
satbai, line 15: It looks like the download URL is different for latest vs other versions. 

Download for latest is: 
`https://github.com/$REPO/releases/latest/download/$FILE` 
and download specific version you need to use the tag name and
`https://github.com/$REPO/releases/download/${version_tag_name}/$FILE`
🤷 

At least I wasn't able to get the first one working with a specific version. 

If you want to get your contribution in, I'm happy to approve and merge this PR after you fix the URL issue if you get it done before Monday. We're releasing a new version of the cred provider on Monday and I'm going to merge my [other PR](https://github.com/microsoft/artifacts-credprovider/pull/302) before doing that. And in that other PR I added the version support as well but I'm happy to merge your PR first if it's ready by the time of the release. :)

satbai, line 15: It looks like the download URL is different for latest vs other versions. 

Download for latest is: 
`https://github.com/$REPO/releases/latest/download/$FILE` 
and download specific version you need to use the tag name and
`https://github.com/$REPO/releases/download/${version_tag_name}/$FILE`
🤷 

At least I wasn't able to get the first one working with a specific version. 

If you want to get your contribution in, I'm happy to approve and merge this PR after you fix the URL issue if you get it done before Monday. We're releasing a new version of the cred provider on Monday and I'm going to merge my [other PR](https://github.com/microsoft/artifacts-credprovider/pull/302) before doing that. And in that other PR I added the version support as well but I'm happy to merge your PR first if it's ready by the time of the release. :)

reillysiemens:
@satbai: Looks like I missed the window on this. Thanks for making sure that was addressed in the v1.0.0 release though. I appreciate this quick responses here.

PR #302, opened 2022-03-04
Add readme and script updates on release 1.0.0

satbai:


helpers/installcredprovider.ps1
```diff
@@ -80,59 +89,65 @@ if (!$releaseId) {
...
-} catch {
-    Write-Error $zipErrorString
-    return
-}
 
-if (!$packageSourceUrl) {
-    Write-Error $zipErrorString
-    return
-}
+function InstallZip {
```
satbai, line 104: I wrapped the download part from the script into a function so that it can be called twice in case the user has -AddNetfx flag and -InstallNet6 flag when calling the script. AddNetfx installs the asset that comes with the netcore3.1 version by default, so after installing the netfx we need to install the net6 one separately. Nothing inside the IstallZip function actually changes, the diff just looks a bit nasty. 

The interesting changes are at the top of the file when we decide the asset name, ($zipFile). And down below where we decide whether we need to download another asset (for net6+netfx scenario).

README.md
```diff
@@ -145,6 +148,12 @@ The Credential Provider accepts a set of environment variables. Not all of them
  {"endpointCredentials": [{"endpoint":"http://example.index.json", "username":"optional", "password":"accesstoken"}]}
 ```
 
+## Release version 1.0.0
+
+Release version [1.0.0](https://github.com/microsoft/artifacts-credprovider/releases/tag/v1.0.0) was released in January 2022. Netcore version 1.0.0 of the Artifacts Credential Provider requires .NET Core 3.1. Older versions than 1.0.0 require .NET Core 2.1.
```
satbai, line 153: Change January to March

README.md
```diff
@@ -70,6 +71,8 @@ Examples:
 
 Using the above is recommended, but as per [NuGet's plugin discovery rules](https://github.com/NuGet/Home/wiki/NuGet-cross-plat-authentication-plugin#plugin-installation-and-discovery), alternatively you can install the credential provider to a location you prefer, and then set the environment variable NUGET_PLUGIN_PATHS to the .dll of the credential provider found in plugins\netcore\CredentialProvider.Microsoft\CredentialProvider.Microsoft.dll. For example, $env:NUGET_PLUGIN_PATHS="my-alternative-location\CredentialProvider.Microsoft.dll".
 
+Users requiring .NET 6 can manually download the .NET 6 version `Microsoft.Net6.NuGet.CredentialProvider` of the [1.0.0 release](https://github.com/microsoft/artifacts-credprovider/releases/tag/v1.0.0).
```
sandyarmstrong, line 74: Might be worth calling out here that ARM64 users will need the .NET 6 version.

satbai: Add install script support as well.

embetten (approved):

NArnott:
Seems like this change just broke my docker builds is Azure Devops:

> Downloading from https://github.com/Microsoft/artifacts-credprovider/releases/download/latest/Microsoft.NuGet.CredentialProvider.tar.gz
gzip: stdin: not in gzip format
tar: Child returned status 1

The correct link is supposed to be https://github.com/Microsoft/artifacts-credprovider/releases/latest/download/Microsoft.NuGet.CredentialProvider.tar.gz

looks like /latest and /download was flipped,

curtstate:
this change broke our docker builds

PR #322, opened 2022-05-25
Implemented error handling + error messaging

AlexVTor:
re-implemented changes on own branch

### DIFF
--- a/helpers/installcredprovider.ps1
+++ b/helpers/installcredprovider.ps1
@@ -19,7 +19,7 @@ param(
     [switch]$InstallNet6
 )
 
-$script:ErrorActionPreference='Stop'
+$script:ErrorActionPreference = 'Stop'
 
 # Without this, System.Net.WebClient.DownloadFile will fail on a client with TLS 1.0/1.1 disabled
 if ([Net.ServicePointManager]::SecurityProtocol.ToString().Split(',').Trim() -notcontains 'Tls12') {
@@ -34,7 +34,8 @@ if ($Version.StartsWith("0.") -and $InstallNet6 -eq $True) {
 $userProfilePath = [System.Environment]::GetFolderPath([System.Environment+SpecialFolder]::UserProfile);
 if ($userProfilePath -ne '') {
     $profilePath = $userProfilePath
-} else {
+}
+else {
     $profilePath = $env:UserProfile
 }
 
@@ -75,7 +76,8 @@ if (![string]::IsNullOrEmpty($Version)) {
         $releaseJson = $releases | ConvertFrom-Json
         $correctReleaseVersion = $releaseJson | ? { $_.name -eq $Version }
         $releaseId = $correctReleaseVersion.id
-    } catch {
+    }
+    catch {
         Write-Error $versionError
         return
     }
@@ -87,7 +89,7 @@ if (!$releaseId) {
 }
 
 $releaseUrl = [System.IO.Path]::Combine($releaseUrlBase, $releaseId)
-$releaseUrl = $releaseUrl.Replace("\","/")
+$releaseUrl = $releaseUrl.Replace("\", "/")
 
 $zipFile = "Microsoft.NetCore3.NuGet.CredentialProvider.zip"
 if ($Version.StartsWith("0.")) {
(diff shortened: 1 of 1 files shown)


helpers/installcredprovider.ps1
```diff
@@ -111,13 +113,20 @@ function InstallZip {
...
         $packageSourceUrl = $zipAsset.browser_download_url
-    } catch {
-        Write-Error $zipErrorString
-        return
+                
+        if (!$releaseJson) {
+            throw("Unable to convert JSON")
+        }
+        if (!$zipAsset) {
+            throw("Unable to retrieve zip asset")
```
satbai, line 121: nit: retrieve sounds like an API call to me, we're just reading the json here so maybe better would be something like "Unable to find asset $zipFile from release json object", and for the packageSourceUrl something like, "Unable to find download url from asset $zipFile"

helpers/installcredprovider.ps1
```diff
@@ -111,13 +113,20 @@ function InstallZip {
...
+        }
+        if (!$packageSourceUrl) {
+            throw("Unable to retrieve source URL")
+        }
     }
-
-    if (!$packageSourceUrl) {
-        Write-Error $zipErrorString
+    catch {
+        $zipErrorString = "$zipErrorString `nError : " + $_.Exception.Message
```
satbai, line 128: Since $zipErrorString is only used in one place anymore, you can delete it from line 109 and just have the error typed here to simplify the code. Would something like this work?
`Write-Error "Unable to resolve the browser download url from $releaseUrl Error: $_.Exception.Message"`

helpers/installcredprovider.ps1
```diff
@@ -34,7 +34,8 @@ if ($Version.StartsWith("0.") -and $InstallNet6 -eq $True) {
 $userProfilePath = [System.Environment]::GetFolderPath([System.Environment+SpecialFolder]::UserProfile);
 if ($userProfilePath -ne '') {
     $profilePath = $userProfilePath
-} else {
+}
+else {
```
satbai, line 38: Is there a reason for the formatting changes?

helpers/installcredprovider.ps1
```diff
@@ -111,13 +113,20 @@ function InstallZip {
         $releaseJson = $release.Content | ConvertFrom-Json
```
satbai, line 113: Doesn't the null check for $releaseJson need to be before we use the variable in line 114? Would line 114 throw without ever getting to the null check below?

satbai: Same with the null check for $zipAsset

satbai: And while we're at it, might as well add a null check for $release as well?

AlexVTor: From my understanding and testing, having a null value for $releaseJon, $zipAsset, or even $packageSourceUrl wont break the script. The following function calls (ConvertFrom-Json / where-object ) will just be made on a null value, resulting in another null return value. So the changes your suggesting would save us a few function calls, but may reduce readability. I do still think that implementing them is a good idea though, since faster run time is never a bad thing. 

Also, I'm not sure if a null check for $release is needed. the Invoke-WebRequest function in line 112 should either be successful and return a WebResponseObject, or throw an error. If I am misunderstanding this or if it will still be beneficial to have, please let me know!

jiaw37: I think it is okay to leave the null check there because if the Invoke-WebRequest does change under the hood or in new version, it would be "safer"

helpers/installcredprovider.ps1
```diff
@@ -104,20 +105,22 @@ if ($AddNetfx -eq $True) {
 function InstallZip {
     Write-Verbose "Using $zipFile"
 
-    $zipErrorString = "Unable to resolve the Credential Provider zip file from $releaseUrl"
     try {
         Write-Host "Fetching release $releaseUrl"
         $release = Invoke-WebRequest -UseBasicParsing $releaseUrl
+        if (!$release) { throw("Unable to make Web Request to $releaseUrl") }
```
jiaw37, line 111:
```suggestion
        if (!$release) { throw ("Unable to make Web Request to $releaseUrl") }
```
I think you are missing a space, same for everywhere else

helpers/installcredprovider.ps1
```diff
@@ -104,20 +105,22 @@ if ($AddNetfx -eq $True) {
...
+        if (!$releaseJson) { throw("Unable to get content from JSON") }
+
         $zipAsset = $releaseJson.assets | ? { $_.name -eq $zipFile }
+        if (!$zipAsset) { throw("Unable to find asset $zipFile from release json object") }  
+
         $packageSourceUrl = $zipAsset.browser_download_url
-    } catch {
-        Write-Error $zipErrorString
-        return
+        if (!$packageSourceUrl) { throw("Unable to find download url from asset $zipFile") }
```
jiaw37, line 120:
```suggestion
        if (!$packageSourceUrl) { throw("Unable to find download url from asset $packageSourceUrl") }
```

AlexVTor: I think we are both mistaken and the correct var to show will be $zipAsset since that's what we are searching through!

satbai (approved):
lgtm

PR #324, opened 2022-06-12
Add support for WAM broker

johnterickson:
So far, I've just tested on Windows.


CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
```diff
@@ -79,29 +82,55 @@ public async Task<IMsalToken> AcquireTokenWithDeviceFlowAsync(Func<DeviceCodeRes
...
+        {
+            IPublicClientApplication publicClient = null;
             try
             {
+                publicClient = await GetPCAAsync(withBroker, useLocalHost: false).ConfigureAwait(false);
+                var accounts = await publicClient.GetAccountsAsync();
+
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    accounts = accounts.Concat(new [] { PublicClientApplication.OperatingSystemAccount});
```
bgavrilMS, line 105: if broker is not enabled, you will not be able to use this special account.

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
```diff
@@ -79,29 +82,55 @@ public async Task<IMsalToken> AcquireTokenWithDeviceFlowAsync(Func<DeviceCodeRes
...
+                    }
                 }
             }
             finally
             {
                 var helper = await GetMsalCacheHelperAsync();
-                helper?.UnregisterCache(publicClient.UserTokenCache);
+                if (publicClient != null)
+                {
+                    helper?.UnregisterCache(publicClient.UserTokenCache);
```
bgavrilMS, line 132: Not needed. You can double register the cache, there won't be a problem.

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
```diff
@@ -79,29 +82,55 @@ public async Task<IMsalToken> AcquireTokenWithDeviceFlowAsync(Func<DeviceCodeRes
...
             try
             {
+                publicClient = await GetPCAAsync(withBroker, useLocalHost: false).ConfigureAwait(false);
+                var accounts = await publicClient.GetAccountsAsync();
+
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    accounts = accounts.Concat(new [] { PublicClientApplication.OperatingSystemAccount});
+                }
+
```
bgavrilMS, line 107: Is there an option for the user to switch account? Beccause what will happen today if the broker is enabled is: 

- user logs in with their Windows account, let's say joe@microsoft.com
- GetAccounts now returns an account for "joe@microsoft.com"
- silent flow always logs in "joe@microsoft.com"


- next time

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
```diff
@@ -79,29 +82,55 @@ public async Task<IMsalToken> AcquireTokenWithDeviceFlowAsync(Func<DeviceCodeRes
...
-                    { }
-                    catch (MsalServiceException)
-                    { }
+                    catch (MsalUiRequiredException e)
+                    { 
+                        this.Logger.Verbose(e.Message);
+                    }
+                    catch (MsalServiceException e)
+                    {
+                        this.Logger.Verbose(e.Message);
```
bgavrilMS, line 123: This should be logged at Error level.

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
bgavrilMS: I think @kyle-rader has this flow better structured in the CLI project and can spot some problems. 
This also shows how to use this concept : https://github.com/bgavrilMS/MSAL-SingleAccount-SSO/blob/master/SSO/Program.cs

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
```diff
@@ -166,29 +176,45 @@ public async Task<IMsalToken> AcquireTokenWithWindowsIntegratedAuth(Cancellation
...
             var helper = await GetMsalCacheHelperAsync().ConfigureAwait(false);
 
             var publicClientBuilder = PublicClientApplicationBuilder.Create(this.clientId)
-                .WithAuthority(this.authority);
+                .WithAuthority(this.authority)
+                .WithLogging(
+                    (LogLevel level, string message, bool containsPii) => {
+                        if (containsPii)
+                        {
+                            this.Logger.Verbose($"MSAL Log ({level}): [REDACTED FOR PII]");
```
jmyersmsft, line 191: Consider a mechanism (e.g. env. variable) to allow logging messages that contain PII

CredentialProvider.Microsoft/Util/EnvUtil.cs
```diff
@@ -76,14 +83,20 @@ public static string GetMsalCacheLocation()
 
         internal static bool MsalEnabled()
         {
-            return GetEnabledFromEnvironment(MsalEnabledEnvVar, defaultValue: false);
+            return GetEnabledFromEnvironment(MsalEnabledEnvVar, defaultValue: true);
```
jmyersmsft, line 86: This cannot default to true unless and until it does not regress the experience for MSA users

Build.props
```diff
@@ -2,5 +2,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <CredentialProviderVersion>1.0.0</CredentialProviderVersion>
+    <IsWindowsHost>$([MSBuild]::IsOSPlatform('Windows'))</IsWindowsHost>
```
johnterickson, line 5: I don't think this is needed anymore

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
kyle-rader: In https://aka.ms/azureauth we currently assume the tool is in use by "corp" accounts, not MSAs, and so for the broker auth mode we will try to use the default OS account if one isn't found first from the cache. 
https://github.com/AzureAD/microsoft-authentication-cli/blob/e3faf814d21441c4353461b23c001382f6dd79f8/src/MSALWrapper/AuthFlow/Broker.cs#L68

kyle-rader: To Support MSA, this likely needs to change though.

CredentialProvider.Microsoft/CredentialProvider.Microsoft.csproj
```diff
@@ -3,37 +3,35 @@
...
     <DefineConstants>TRACE</DefineConstants>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Identity.Client" Version="4.32.1" />
-    <PackageReference Include="Microsoft.Identity.Client.Extensions.Msal" Version="2.18.5" />
+    <PackageReference Include="Microsoft.Identity.Client" Version="4.45.0"/>
+    <PackageReference Include="Microsoft.Identity.Client.NativeInterop" Version="0.8.0" Condition=" '$(TargetFramework)' == 'net461' "/>
```
bgavrilMS, line 32: Hmm, is this required? I was under the impression that the ref to `Microsoft.Identity.Client.Broker` is sufficient.

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
```diff
@@ -43,10 +50,13 @@ private async Task<MsalCacheHelper> GetMsalCacheHelperAsync()
...
             if (helper == null && this.cacheEnabled && RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
+                this.Logger.Verbose($"Using MSAL cache at `{cacheLocation}`.");
+
                 var fileName = Path.GetFileName(cacheLocation);
                 var directory = Path.GetDirectoryName(cacheLocation);
 
-                var builder = new StorageCreationPropertiesBuilder(fileName, directory, this.clientId);
+                var builder = new StorageCreationPropertiesBuilder(fileName, directory);
+                builder = builder.WithCacheChangedEvent(this.clientId, "https://login.microsoftonline.com/common");
```
bgavrilMS, line 59: Is this really needed? This allows you to subscribe to an event which tells you if the cache file has changed. I do not see an event handler for this though.

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
bgavrilMS, line 191: You don't need to do this, you can just set "use PII" to false - this is an optional param of this method. 

Also, have a look at the overload with `IIDentityLogger`, which is easier to control.

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
johnterickson: @jmyersmsft thoughts on how to approach this?

CredentialProvider.Microsoft/Util/EnvUtil.cs
johnterickson: ah that was just from testing - good catch!

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
jmyersmsft: I honestly have no idea. Everything about authentication is always as painful as possible IME

kyle-rader: In AzureAuth we have a slightly more narrow use case where we expect your machine (if windows) is using a corp-conected account/machine (either AD or AAD), and so if no cached account is found, we automatically try using the Default OS account. We also provide a suffix (domain) filter, to filter cached accounts, and if there's only 1 account after filtering, we will use that - all to avoid account pickers. This is in https://github.com/AzureAD/microsoft-authentication-cli/blob/2836f99cfd1eeb62814056b322765bdcf85a6e86/src/MSALWrapper/PCAWrapper.cs#L117

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
kyle-rader: We do not listen for cache change events in AzureAuth. I thought this was designed for long-running applications like VS.

CredentialProvider.Microsoft/CredentialProvider.Microsoft.csproj
johnterickson: For whatever reason, the transitive dependency was not getting pulled in for net461.  I will double check.

johnterickson: @bgavrilMS Confirmed.  When I comment this out and run `test.bat` I get:

```
[Verbose] [CredentialProvider]MSAL Log (Error): False MSAL 4.46.0.0 MSAL.Desktop 4.8 or later Windows 10 Enterprise [2022-08-05 15:14:11Z - c7aa1236-f1c7-45c8-8c03-8269c30990df] Exception type: System.DllNotFoundException

   at Microsoft.Identity.Client.NativeInterop.API.x64.MSALRUNTIME_Startup()
   at Microsoft.Identity.Client.NativeInterop.API.x64.Startup()
   at Microsoft.Identity.Client.NativeInterop.Module.AddRef()
   at Microsoft.Identity.Client.NativeInterop.Core..ctor()
   at Microsoft.Identity.Client.Broker.RuntimeBroker.<AcquireTokenSilentDefaultUserAsync>d__12.MoveNext()

Comment (johnterickson) [reply CredentialProvider.Microsoft/CredentialProvider.Microsoft.csproj:32]: net6.0 and netcoreapp3.1 work fine 🤷‍♂️

File: CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs (hunk 9 above)
Comment (johnterickson) [reply CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs:59]: Removed

File: CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs (hunk 3 above)
Comment (johnterickson) [reply CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs:95]: I wonder if I can compare the cached account tenant to the authority...

Comment (johnterickson) [reply CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs:96]: That worked well!

File: CredentialProvider.Microsoft/CredentialProviders/Vsts/IAuthUtil.cs [hunk 10]
```diff
@@ -81,7 +84,7 @@ public async Task<Uri> GetAadAuthorityUriAsync(Uri uri, CancellationToken cancel
             var aadBase = UsePpeAadUrl(uri) ? "https://login.windows-ppe.net" : "https://login.microsoftonline.com";
             logger.Verbose(string.Format(Resources.AADAuthorityNotFound, aadBase));
 
-            var tenant = EnvUtil.MsalEnabled() ? OrganizationsTenant: CommonTenant;
+            var tenant = EnvUtil.MsalEnabled() ? MsaTenant.ToString() : CommonTenant;
```
johnterickson, line 87: Is it safe to assume this is MSA?

Build.props
```diff
@@ -2,6 +2,25 @@
...
     <CredentialProviderVersion>1.0.0</CredentialProviderVersion>
-    <TargetFrameworks>netcoreapp3.1;net461;net6.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net461;net6.0;net6.0-windows10.0.17763.0</TargetFrameworks>
+    <Broker>NONE</Broker>
+  </PropertyGroup>
+  
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <Broker>NONE</Broker>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0-windows10.0.17763.0'">
```
johnterickson, line 12: need to change the install scripts for this, too

CredentialProvider.Microsoft/Program.cs
```diff
@@ -22,6 +22,7 @@
 using PowerArgs;
 using ILogger = NuGetCredentialProvider.Logging.ILogger;
 
+// [assembly: InternalsVisibleToAttribute("CredentialProvider.Microsoft.Tests")]
```
johnterickson, line 25: remove

helpers/installcredprovider.sh
```diff
@@ -16,8 +16,13 @@ NUGET_PLUGIN_DIR="$HOME/.nuget/plugins"
 
 # determine whether we install default or Net6
 if [[ ! -z ${USE_NET6_ARTIFACTS_CREDENTIAL_PROVIDER} ]]; then
-  FILE="Microsoft.Net6.NuGet.CredentialProvider.tar.gz"
-
+  if [[ -z "WINDIR" ]]
```
jmyersmsft, line 19: I think you mean `[[ ! -v "WINDIR" ]]` or `[[ -z "${WINDIR}" ]]`

helpers/installcredprovider.sh
```diff
@@ -16,8 +16,13 @@ NUGET_PLUGIN_DIR="$HOME/.nuget/plugins"
...
+  then
+    FILE="Microsoft.Net6.NuGet.CredentialProvider.tar.gz"
+  else
+    FILE="Microsoft.Net6-Windows.NuGet.CredentialProvider.tar.gz"
```
johnterickson, line 23: @jmyersmsft It looks like `https://aka.ms/install-artifacts-credprovider.sh` redirects to `https://raw.githubusercontent.com/microsoft/artifacts-credprovider/master/helpers/installcredprovider.sh`.

This is bad because an official `Microsoft.Net6-Windows.NuGet.CredentialProvider.tar.gz` does not exist yet and so as soon as this merges, this will break:
```
REDMOND+jerick@jerick-hpz440 MINGW64 /d/src/artifacts-credprovider/helpers (broker)
$ USE_NET6_ARTIFACTS_CREDENTIAL_PROVIDER=1 ./installcredprovider.sh
Downloading from https://github.com/Microsoft/artifacts-credprovider/releases/latest/download/Microsoft.Net6-Windows.NuGet.CredentialProvider.tar.gz

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

I guess this installer changes need to come later?

jmyersmsft: oh yeah, forgot about that

johnterickson: ok commented out for now

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
```diff
@@ -62,46 +71,105 @@ public async Task<IMsalToken> AcquireTokenWithDeviceFlowAsync(Func<DeviceCodeRes
...
             }
-            finally
+
+            foreach ((IAccount account, string canonicalName, _) in sortedAccounts)
             {
-                var helper = await GetMsalCacheHelperAsync();
-                helper?.UnregisterCache(publicClient.UserTokenCache);
+                try
+                {
+                    this.Logger.Verbose($"Attempting to use identity `{canonicalName}`.");
```
johnterickson, line 160: Does this actually reach out to AzDO to see if the identity has access?

CredentialProvider.Microsoft/CredentialProvider.Microsoft.csproj
```diff
@@ -19,17 +19,19 @@
...
+    <DefineConstants>$(DefineConstants);TRACE</DefineConstants>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Identity.Client" Version="4.46.2" />
-    <PackageReference Include="Microsoft.Identity.Client.Extensions.Msal" Version="2.18.5" />
+    <PackageReference Include="Microsoft.Identity.Client" Version="4.47.1" />
+    <PackageReference Include="Microsoft.Identity.Client.Broker" Version="4.47.1-preview"/>
+    <PackageReference Include="Microsoft.Identity.Client.NativeInterop" Version="0.*" Condition=" '$(TargetFramework)' == 'net461' "/>
+    <PackageReference Include="Microsoft.Identity.Client.Extensions.Msal" Version="2.23.0" />
```
johnterickson, line 34: bump to latest

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
```diff
@@ -166,32 +229,88 @@ public async Task<IMsalToken> AcquireTokenWithWindowsIntegratedAuth(Cancellation
...
+                    publicClientBuilder = publicClientBuilder.WithBrokerPreview();
+
+                    //  needed for WAM
+                    publicClientBuilder = publicClientBuilder.WithParentActivityOrWindow(() => GetConsoleOrTerminalWindow());
+
+                    publicClientBuilder = publicClientBuilder.WithWindowsBrokerOptions(new WindowsBrokerOptions() {
+                        HeaderText = "Azure DevOps Artifacts",
+                        // System.NotImplementedException: The new broker implementation does not yet support Windows account discovery (ListWindowsWorkAndSchoolAccounts option)
+                        // https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3602
+                        // ListWindowsWorkAndSchoolAccounts = true,
```
Darsstar, line 263: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3602 has been merged last month. The first release is [4.49.0](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.49.0), 4.49.1 is currently the latest.

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs
JohnSchmeichel: If you're referring to the MSAL acquire token then, no, it won't. Only when that's used. If there are multiple identities found that could reasonably connect (like corp + alt account, multiple MSAs, etc.) then we'd either need to try and use them, or have the user provide the logon hint in those cases. Could image there would be cases where both have "access", but really only one has the right set of permissions for the specific feed being used.

CredentialProvider.Microsoft/CredentialProviders/Vsts/IAuthUtil.cs
JohnSchmeichel: I believe that's what the "organizations" was doing before, but will confirm

CredentialProvider.Microsoft/CredentialProviders/Vsts/AdalTokenProvider.cs
```diff
@@ -74,11 +74,11 @@ public async Task<IAdalToken> AcquireTokenSilentlyAsync(CancellationToken cancel
...
                 return null;
             }
-        }
-
-#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
-        public async Task<IAdalToken> AcquireTokenWithUI(CancellationToken cancellationToken)
-#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
+        }
+
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
```
bgavrilMS, line 79: nit: there is no need to disable this warning. Just remove the `async` keyword. It won't affect inheritance.

CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalBearerTokenProvidersFactory.cs
```diff
@@ -12,22 +14,26 @@ internal class MsalBearerTokenProvidersFactory : IBearerTokenProvidersFactory
...
+
+            foreach(bool brokerEnabled in options)
             {
-                new MsalCacheBearerTokenProvider(msalTokenProvider),
-                new MsalWindowsIntegratedAuthBearerTokenProvider(msalTokenProvider),
-                new MsalUserInterfaceBearerTokenProvider(msalTokenProvider),
-                new MsalDeviceCodeFlowBearerTokenProvider(msalTokenProvider)
-            };
+                IMsalTokenProvider msalTokenProvider = msalTokenProviderFactory.Get(authority, brokerEnabled, logger);
+                yield return new MsalSilentBearerTokenProvider(msalTokenProvider);
```
bgavrilMS, line 32: Does this also attempt to log in with the default OS account? It should work on Windows.

JohnSchmeichel: Yes it does: https://github.com/microsoft/artifacts-credprovider/blob/9671beb3c3ed537e04678a9bf36ada2f4777667d/CredentialProvider.Microsoft/CredentialProviders/Vsts/MSAL/MsalTokenProvider.cs#L134

Currently used as a fallback if there are no usable cached accounts

CredentialProvider.Microsoft/CredentialProviders/Vsts/AdalTokenProvider.cs
JohnSchmeichel: This is because the method has pragmas between .NET Framework and .NET Core, one of which has await and the other doesn't. Alternatively the entire method could be pragma'd instead of just the impl. Hoping to remove all this once MSAL is rolled out and made the default.

CredentialProvider.Microsoft/CredentialProviders/Vsts/IAuthUtil.cs
JohnSchmeichel: 'organizations' includes work and school accounts, and MSA users if passthrough is enabled. If it was a work or school account then there would have been a tenant id available from the unauthenticated request against the Azure DevOps endpoint, which leaves only MSA passthrough users.

JohnSchmeichel:
> So far, I've just tested on Windows.

Tested on Linux now as well.

JohnSchmeichel (approved):

ShadowXVII:
Woo, this works well after setting NUGET_CREDENTIALPROVIDER_MSAL_ENABLED and NUGET_CREDENTIALPROVIDER_MSAL_ALLOW_BROKER. Looking forward to this being pushed to release.

PR #334, opened 2022-08-30
Add PAT lifetime fallback support

johnterickson:
AzDO service now has a policy that will restrict PAT duration.


johnterickson:
This is a subset of #324

jmyersmsft (approved):

PR #354, opened 2022-11-17
Fix credential provider release build and resolve build warnings

JohnSchmeichel:
The release build for ingestion to Visual Studio was broken by service connection remediation so needed to be recreated. Some recent changes in the build templates are now no longer compatible with the build machines running release builds. Also fixing some miscellaneous build and release warnings that have snuck in.

### DIFF
--- a/build/build.yml
+++ b/build/build.yml
@@ -14,7 +14,10 @@ steps:
   inputs:
     packageType: sdk
     version: 6.x
-    
+
+- task: NuGetToolInstaller@1
+  displayName: NuGet Install
+
 - task: NuGetAuthenticate@1
   displayName: NuGet Authenticate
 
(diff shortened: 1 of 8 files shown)


build/build.yml
```diff
@@ -30,39 +30,47 @@ steps:
     inputs:
       feedSource: 'https://pkgs.dev.azure.com/mseng/_packaging/MicroBuildToolset/nuget/v3/index.json'
 
-  - script: nuget restore CredentialProvider.Microsoft.VSIX/packages.config -OutputDirectory packages
+  - task: NuGetCommand@2
```
ByAgenT, line 33: using `nuget restore` in script is actually recommended way to connect to nuget in pipelines :) [NuGet command task](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/package/nuget?view=azure-devops) is slowly being deprecated. The main difference between those two approaches is that NuGetCommand is isolated and does not respect NuGet Authenticate task nor it shares auth with other tasks and steps, which is more inconvenient. Examples in [NuGet Authenticate documentation](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/package/nuget-authenticate?view=azure-devops#nugetexe-1) is recommended practice now
image

ByAgenT (approved):
Looks good to me overall, left one suggestion for the nuget restore step

build/build.yml
JohnSchmeichel: Ah, interesting... main reason moving away was that the release builds don't run on Azure Pipelines but run on an approved release pool for signing, and this either doesn't have nuget pre-installed, or not on PATH, so straight nuget calls fail.

Maybe an option we can try here is to use dotnet restore or dotnet nuget to pick up the right nuget location.

ByAgenT: I think it just misses [NugetTool Installer](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/nuget?view=azure-devops) step

ByAgenT: But `dotnet restore` should be fine as well I think

JohnSchmeichel: checked the docs (or rather the GitHub issue where it was discussed), dotnet restore doesn't support package.config, only csproj PackageReferences, so need to use nuget outside of dotnet tooling. Trying out NuGet installer, which should be a no-op in most cases.

JohnSchmeichel: NuGet installer (and fixing the command) worked!

PR #388, opened 2023-04-18
Create Microsoft.Artifacts.Authentication package

JohnSchmeichel:
Proposed changes to pull the MSAL related types into their own project and deliver as an extension to MSAL. Design and implementation are a work in progress to collect feedback on the approach and API design.

Example usage of to create a PublicClientApplication with recommended settings and defaults for Azure Artifacts and enumerate providers:

```csharp
var app = AzureArtifacts.CreateDefaultBuilder(authority)
    .WithBroker(true, logger)
    .WithLogging((LogLevel level, string message, bool containsPii) =>
    {
        // Application specific logging
    })
    .Build();

// Can use MsalTokenProviders which works for most cases, or compose the token providers manually
var providers = MsalTokenProviders.Get(app, logger);

var tokenRequest = new TokenRequest("https://pkgs.dev.azure.com/org")
{
    IsInteractive = true
};

foreach (var provider in providers)
{
    if (!provider.CanGetToken(tokenRequest))
        continue;

    var result = await provider.GetTokenAsync(tokenRequest);
}
```

1. Uses and extends the MSAL token types, and not trying to be a wrapper around them. No plans to support more token authentication libraries in the future.
2. Uses Microsoft.Extensions.Logging as the logging infrastructure, and adaptors will be created where necessary to interface with client tooling.
3. All settings and defaults are pulled out onto TokenRequest, or are composable using extension methods.


src/Authentication/MsalCache.cs
```diff
@@ -0,0 +1,96 @@
...
+            logger.LogWarning("warning: cannot persist Microsoft authentication token cache securely!");
+            logger.LogTrace(ex.ToString());
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                // On macOS sometimes the Keychain returns the "errSecAuthFailed" error - we don't know why
+                // but it appears to be something to do with not being able to access the keychain.
+                // Locking and unlocking (or restarting) often fixes this.
+                logger.LogError(
+                    "warning: there is a problem accessing the login Keychain - either manually lock and unlock the " +
```
embetten, line 60: Do we need to do localization on these logging message ?  Are we doing localization at all currently in the repo?

JohnSchmeichel: We probably should, but are not currently. The rest of the credential provider (except the MSAL stuff and some ad-hoc places) uses resource files, but we're not building or localizing this at all currently.

JohnSchmeichel: Went on a journey and ended up re-working how resource generation is done entirely, and added resource files for the new library.

src/Authentication/AzureArtifacts.cs
```diff
@@ -0,0 +1,87 @@
...
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.Broker;
+
+namespace Microsoft.Artifacts.Authentication;
+
+public static class AzureArtifacts
+{
+    public const string ClientId = "d5a56ea4-7369-46b8-a538-c370805301bf";
+
+    public static PublicClientApplicationBuilder CreateDefaultBuilder(Uri authority, ILogger logger)
```
embetten, line 16: The logger is not used in the method - do we need it here?

JohnSchmeichel: Well, we used to need it, which I realize is not a great answer. Will remove.

embetten: reminder to remove from readme example to match as well

embetten (approved):

PR #395, opened 2023-04-19
Conda cred provider wrapper

zhenghao104:
A credential provider wrapper for Conda protocol. It will call the credential provider and set the token into an environment variable for Conda client. The conda package artifacts-credprovider-conda will be published to anaconda.org for users to consume.


JohnSchmeichel:
> > 
> 
> If we are restructuring the directory, it can be a separate PR before/after this PR to keep this PR clean.

I wouldn't do a big restructure PR, but rather start to move things into place over time. Think it's reasonable to say new code goes into the "right" place to start so we don't need to go back and fix it up later.

zhenghao104:
Off topic but we should also create an automated process that will publish the cred stuff (vsts-npm-auth/artifacts-keyring) to the respective public repositories. They are all done manually now through command prompt.

JohnSchmeichel:
> Off topic but we should also create an automated process that will publish the cred stuff (vsts-npm-auth/artifacts-keyring) to the respective public repositories. They are all done manually now through command prompt.

Agreed! Have this feature in the backlog to take care of that: https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2048155

zhenghao104:
> > Off topic but we should also create an automated process that will publish the cred stuff (vsts-npm-auth/artifacts-keyring) to the respective public repositories. They are all done manually now through command prompt.
> 
> Agreed! Have this feature in the backlog to take care of that: https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2048155

For the first iteration, I am going to do it the same way as vsts-npm-auth and artifacts keyring by creating our own account.

JohnSchmeichel (approved):

src/artifacts-credprovider-conda/src/activate/artifacts-cred-activate.bat
```diff
@@ -0,0 +1,4 @@
+echo Running artifacts-cred-provider-conda...
+set NUGET_CREDENTIALPROVIDER_MSAL_ENABLED=true
+for /f %%i in ('conda config --show --json ^| %CONDA_PYTHON_EXE% %~dp0%artifacts-cred.py') do "set ARTIFACTS_CONDA_TOKEN=%%i"
```
zhenghao104, line 3: This does not work, not sure why yet

src/artifacts-credprovider-conda/src/activate/artifacts-cred-activate.bat
```diff
@@ -0,0 +1,4 @@
+echo Running artifacts-cred-provider-conda...
```
jmyersmsft, line 1: Let's try to avoid writing text to stdout in these scripts. Also, for the .bat ones, add `@echo off` to the top so they don't print out the commands as they get executed

jmyersmsft (changes requested):

src/artifacts-credprovider-conda/src/activate/artifacts-cred-activate.bat
zhenghao104: Removed print statement from all scripts

jmyersmsft (approved):

PR #397, opened 2023-04-24
Cache http response headers

JohnSchmeichel:
All of the methods off of IAuthUtil will probe the same url to get response headers. We were making 3 requests for the same response, and after #392 we made one less in the common case. Adding an internal cache to simply cache the response headers by url. It's possible to refactor to make this a little more explicit (e.g. pull the headers once and pass that around) but those changes ended up being quite large and may cause more work with planned changes.

Also passing the HttpCompletionOption.ResponseHeadersRead option since we don't care about the request body or any trailing headers.

### DIFF
--- a/CredentialProvider.Microsoft/CredentialProviders/Vsts/IAuthUtil.cs
+++ b/CredentialProvider.Microsoft/CredentialProviders/Vsts/IAuthUtil.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
 using System.Net.Http.Headers;
@@ -39,6 +40,7 @@ namespace NuGetCredentialProvider.CredentialProviders.Vsts
         public static readonly Guid FirstPartyTenant = Guid.Parse("f8cdef31-a31e-4b4a-93e4-5f571e91255a");
         public static readonly Guid MsaAccountTenant = Guid.Parse("9188040d-6c67-4c5b-b112-36a304b66dad");
 
+        private readonly Dictionary<Uri, HttpResponseHeaders> cache = new();
         private readonly ILogger logger;
 
         public AuthUtil(ILogger logger)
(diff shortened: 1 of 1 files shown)


embetten (approved):

CredentialProvider.Microsoft/CredentialProviders/Vsts/IAuthUtil.cs
```diff
@@ -145,10 +152,10 @@ protected virtual async Task<HttpResponseHeaders> GetResponseHeadersAsync(Uri ur
                     httpClient.DefaultRequestHeaders.UserAgent.Add(userAgent);
                 }
 
-
                 logger.Verbose($"GET {uri}");
-                using (var response = await httpClient.SendAsync(request, cancellationToken))
+                using (var response = await httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken))
```
johnterickson, line 156: I'm not sure if reading only the headers and then disposing the response will make the connection non-reusable.  If this causes the connection to close when it would otherwise be re-used later (with auth), then it might make sense to just read the whole thing here.

johnterickson: oh goodness and we were newing up a new HttpClient each time :(

johnterickson (approved):

CredentialProvider.Microsoft/CredentialProviders/Vsts/IAuthUtil.cs
JohnSchmeichel: Yes yes, HttpClient usage is quite terrible and would love to fix that... can't spend all my time refactoring the code ;). Will be slowly morphing this into the modern tech world, and using IHttpClientFactory and the like is part of that.

JohnSchmeichel: > I'm not sure if reading only the headers and then disposing the response will make the connection non-reusable. If this causes the connection to close when it would otherwise be re-used later (with auth), then it might make sense to just read the whole thing here.

We would never reuse anyway, we never actually hit the pkgs endpoint for anything else, only authentication endpoints. And for creating session tokens that's a different host.

PR #398, opened 2023-04-24
Add devcontainer.json

JohnSchmeichel:
Add default devcontainer.json to enable the repository to build in Codespaces. This builds from the dotnet 6.0 images and adds .NET Core 3.1. The repository is able to build, run, and test if using .NET 6.0 (probably .NET Core 3.1 as well but untested). .NET Framework requires mono to run but didn't see any convenient features for that.

The default devcontainer image contains .NET 7.0 but the container is unable to pull or build by default if the latest SDKs are not already present in the artifacts feed. Since only a single version of .NET can be installed using their feature support, going with the changes here. Once .NET Core 3.1 is dropped we can switch back to the default image + a .NET 6.0 feature.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,5 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/dotnet:6.0",
+  "features": {
+  }
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,5 +1,8 @@
 {
   "image": "mcr.microsoft.com/devcontainers/dotnet:6.0",
   "features": {
+    "ghcr.io/devcontainers/features/dotnet:1": {
+      "version": "3.1"
+    }
   }
 }


johnterickson:
@markphip Was there something you hacked together to get VS Code creds passed through into the container? If so, this could be a code place to showcase it.

johnterickson (approved):

markphip:
If you want to try it, see this:

https://github.com/microsoft/codespace-features/tree/main/src/artifacts-helper

Just adding that feature to your features in devcontainer.json would take care of the nuget authentication in the Codespace. User will get a browser prompt to authenticate to AzDO. The feature installs artifact credential helper in the Codespace and sets up bash alias for dotnet and nuget to get a token and set envvar before running the dotnet or nuget commands.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,11 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/dotnet",
+  "features": {
+    "ghcr.io/devcontainers/features/dotnet:1": {
```
markphip, line 4: Does adding the dotnet feature bring any benefit when you are already using the dotnet image? If you are not using prebuilds, it probably makes it just a bit slower to start a new Codespace to have this feature specified.

JohnSchmeichel: This is mainly for external users... they'd only be able to pull packages that are already present in the Azure DevOps feed, so if .NET 6.0 releases an update and one of us hasn't pulled that version in (explicitly or via pipelines) then they'd be unable to restore. Their workaround would be to manually install the .NET 6.0 SDK, or do some NuGet version pinning. 

Unfortunately the way dotnet and NuGet work they will determine what is the "latest" and want to use it, even if a lower version is present and would work. If a version of .NET 6.0 SDK is already installed then this is avoided. This is why my original image was .NET 6.0 and explicitly brought in .NET Core 3.1. The same issue could occur for .NET 3.1 as well, however it's out of support and the latest version is already pulled in (and we're removing it from the repository).

PR #400, opened 2023-04-26
Add http client factory and user agent to identity provider requests

JohnSchmeichel:
When using MSAL or ADAL authentication libraries no attempt is done to populate the UserAgent header when making requests, and no great API is provided to easily set this. The only mechanism is to implement their IHttpClientFactory interface and customize the HttpClients being used by the library. As such I'm taking this opportunity to use this factory to return a singleton instance for use in all places that we're making HTTP requests.

Additionally I'm updating the UserAgent header we're sending to include more information about the platform and runtime, which now produces a header that looks like the following: ```User-Agent: CredentialProvider.Microsoft/1.0.4+aae4981de95d543b7935811c05474e393dd9e144 (Windows; X64; Microsoft Windows 10.0.19045) CLR/6.0.16 (.NETCoreApp,Version=v6.0; win10-x64; .NET 6.0.16)```

### DIFF
--- a/CredentialProvider.Microsoft/Program.cs
+++ b/CredentialProvider.Microsoft/Program.cs
@@ -29,6 +29,8 @@ namespace NuGetCredentialProvider
         internal static string Name => name.Value;
         internal static string Version => version.Value;
 
+        // Produces a value similar to the following:
+        // CredentialProvider.Microsoft/1.0.4+aae4981de95d543b7935811c05474e393dd9e144 (Windows; X64; Microsoft Windows 10.0.19045) CLR/6.0.16 (.NETCoreApp,Version=v6.0; win10-x64; .NET 6.0.16)
         internal static IList<ProductInfoHeaderValue> UserAgent
         {
             get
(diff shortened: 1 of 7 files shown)


johnterickson:
Can you file an issue in MSAL? There should be some default...

CredentialProvider.Microsoft/Program.cs
```diff
@@ -29,30 +29,30 @@ public static class Program
...
         private static Lazy<string> name = new Lazy<string>(() =>
         {
-            return typeof(Program).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyProductAttribute>()?.Product ?? "CredentialProvider.Microsoft";
+            return typeof(Program).Assembly.GetCustomAttribute<AssemblyProductAttribute>()?.Product ?? "CredentialProvider.Microsoft";
         });
 
         private static Lazy<string> version = new Lazy<string>(() =>
         {
-            return typeof(Program).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? "";
+            return typeof(Program).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? "";
```
johnterickson, line 55: I think we want EntryAssembly so that we can push this down into the NuGet package for npm use

johnterickson (approved):

CredentialProvider.Microsoft/Program.cs
JohnSchmeichel: I'll look into this.. will need to do a bit more changes to bring this into the package so planned on doing some cleanup work there

PR #409, opened 2023-05-04
Update installcredprovider.sh

ZerNox:
Remove hardcode


JohnSchmeichel:
Not opposed to the change (it's functionally the same), is there an issue you're hitting here?

PR #414, opened 2023-05-26
Use default credentials on http client handler

JohnSchmeichel:
MSAL integrated Windows authentication requires using default credentials. This is done if using MSALs http client factory, so must explicitly opt into this behavior.

### DIFF
--- a/CredentialProvider.Microsoft/Util/HttpClientFactory.cs
+++ b/CredentialProvider.Microsoft/Util/HttpClientFactory.cs
@@ -2,8 +2,6 @@
 //
 // Licensed under the MIT license.
 
-using System;
-using System.Net;
 using System.Net.Http;
 using Microsoft.Artifacts.Authentication;
 using IAdalHttpClientFactory = Microsoft.IdentityModel.Clients.ActiveDirectory.IHttpClientFactory;
--- a/CredentialProvider.Microsoft/Util/HttpClientFactory.cs
+++ b/CredentialProvider.Microsoft/Util/HttpClientFactory.cs
@@ -29,11 +29,12 @@ namespace NuGetCredentialProvider.Util
                 UseDefaultCredentials = true
             });
 #else
-            var httpClient  = new HttpClient(new SocketsHttpHandler
-            {
-                PooledConnectionLifetime = TimeSpan.FromMinutes(15),
-                DefaultProxyCredentials = CredentialCache.DefaultCredentials
-            });
+            var httpClient = new HttpClient(
+                new SocketsHttpHandler
+                {
+                    PooledConnectionLifetime = TimeSpan.FromMinutes(15),
+                    Credentials = CredentialCache.DefaultCredentials,
+                });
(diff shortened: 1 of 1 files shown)


JohnSchmeichel:
May need to move away from SocketsHttpHandler here as there is different behavior between the two, still testing.

CredentialProvider.Microsoft/Util/HttpClientFactory.cs
```diff
@@ -23,11 +24,15 @@ static HttpClientFactory()
...
+            var httpClient = new HttpClient(new HttpClientHandler
+            {
+                UseDefaultCredentials = true
+            });
 #else
             var httpClient  = new HttpClient(new SocketsHttpHandler
             {
-                PooledConnectionLifetime = TimeSpan.FromMinutes(15)
+                PooledConnectionLifetime = TimeSpan.FromMinutes(15),
+                DefaultProxyCredentials = CredentialCache.DefaultCredentials
```
JohnSchmeichel, line 35: Not a great replacement for UseDefaultCredentials in SocketsHttpHandler it seems, may need to set Credentials here, though need more testing on that change. Can also just switch away from SocketsHttpHandler entirely which is what MSAL has done.

johnterickson: I changed this to just match MSAL and confirmed it works.

johnterickson (approved):
:shipit:

CredentialProvider.Microsoft/Util/HttpClientFactory.cs
```diff
@@ -21,15 +22,12 @@ public HttpClientFactory(HttpClient httpClient) : base(httpClient)
 
         static HttpClientFactory()
         {
-            // https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines
-#if NETFRAMEWORK
-            var httpClient = new HttpClient();
-#else
-            var httpClient  = new HttpClient(new SocketsHttpHandler
+            var httpClient = new HttpClient(new HttpClientHandler
```
bgavrilMS, line 25: Your default HTTP client factory is now pretty much identical to MSAL's. Might as well not provide one. 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs

bgavrilMS (approved):

CredentialProvider.Microsoft/Util/HttpClientFactory.cs
johnterickson: huh you are right. @JohnSchmeichel  I thought we were doing this for user agent info?

johnterickson: ah the magic is in the base class `MsalHttpClientFactory`

PR #430, opened 2023-06-28
Add explicit version for Newtonsoft.Json

JohnSchmeichel:
NuGet brings in very old versions of Newtonsoft.Json. Need to look at upgrading NuGet dependencies, but in the meantime, bring back the explicit version.


embetten (approved):
LGTM

johnterickson:
https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management#transitive-pinning

PR #444, opened 2023-08-04
Use latest version of NuGet.Protocol

JohnSchmeichel:


embetten (commented):
Oh gosh is there a way to automate this or remind us to bump this version. Worried we will get stale again

embetten (approved):

JohnSchmeichel:
> Oh gosh is there a way to automate this or remind us to bump this version. Worried we will get stale again

Yeah think I might know a service that handles that ;)

PR #469, opened 2024-01-31
Changing away from MicroBuildUploadVstsDropFolder@2

amp-powell:
Using output method with microBuildVstsDrop instead of MicroBuildUploadVstsDropFolder@2 task for compatibility

### DIFF
--- a/build/build.yml
+++ b/build/build.yml
@@ -213,5 +213,6 @@ steps:
   - task: PublishPipelineArtifact@1
     displayName: Publish Artifact $(Build.BuildNumber)
     inputs:
-      PathtoPublish: '$(Build.ArtifactStagingDirectory)\$(BuildConfiguration)\'
-      ArtifactName: '$(Build.BuildNumber)'
+      artifact: '$(Build.BuildNumber)'
+      targetPath: '$(Build.ArtifactStagingDirectory)\$(BuildConfiguration)\'
+
--- a/build/build.yml
+++ b/build/build.yml
@@ -219,9 +219,3 @@ steps:
     inputs:
       BuildDropPath: '$(Build.ArtifactStagingDirectory)\$(BuildConfiguration)\'
       Verbosity: Verbose
-
-  - task: PublishBuildArtifacts@1
-    displayName: Publish Artifact $(Build.BuildNumber)
-    inputs:
-      PathtoPublish: '$(Build.ArtifactStagingDirectory)\$(BuildConfiguration)\'
-      ArtifactName: '$(Build.BuildNumber)'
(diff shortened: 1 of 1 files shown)


JohnSchmeichel (changes requested):

build/build.yml
```diff
@@ -63,13 +63,6 @@ steps:
       configuration: '$(BuildConfiguration)'
       msbuildArguments: '/p:OutputPath=$(Build.ArtifactStagingDirectory)\$(BuildConfiguration)\vsix\ /p:TreatWarningsAsErrors=true'
 
-  - task: ms-vseng.MicroBuildTasks.4305a8de-ba66-4d8b-b2d1-0dc4ecbbf5e8.MicroBuildUploadVstsDropFolder@2
```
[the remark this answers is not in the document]
amp-powell: Changed this!

build/build.yml
```diff
@@ -210,13 +192,6 @@ steps:
       ARTIFACTSERVICES_SYMBOL_ACCOUNTNAME: 'microsoft'
       ARTIFACTSERVICES_SYMBOL_USEAAD: 'false'
 
-- ${{ if eq(parameters.publish, 'true') }}:
```
JohnSchmeichel, line 213: Let's keep this section for uploading symbols if we can, for the PR/CI pipeline we want symbols uploaded. When running from the internal pipeline publish is set to false so this would be avoided

JohnSchmeichel (changes requested):

build/build.yml
```diff
@@ -233,9 +219,3 @@ steps:
     inputs:
```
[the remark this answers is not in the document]
amp-powell: Yes, I'm using 1es pt set ups for SBOM

build/build.yml
amp-powell: Do 1es templates direct us away from using PublishPipelineArtifact@1?

amp-powell: Added this back!

JohnSchmeichel (approved):

build/build.yml
```diff
@@ -227,15 +209,3 @@ steps:
...
     displayName: dotnet push
     condition: eq(variables['PushNupkg'], 'true')
-
-  - task: ManifestGeneratorTask@0
-    displayName: 'SBOM Generator'
-    inputs:
-      BuildDropPath: '$(Build.ArtifactStagingDirectory)\$(BuildConfiguration)\'
-      Verbosity: Verbose
-
-  - task: PublishBuildArtifacts@1
```
JohnSchmeichel, line 237: Same comment here, let's switch this to PublishPipelineArtifact and keep for the PR/CI case

amp-powell: Added this back!

JohnSchmeichel (changes requested):
Double check the PR pipeline is working correctly

JohnSchmeichel (approved):

PR #470, opened 2024-02-06
Enable cross platform session token cache global locking

### LINKED ISSUE #416: Session token cache global lock doesn't work cross-platform
As reported by https://github.com/microsoft/artifacts-keyring/issues/59 the global lock used by the session token cache doesn't work cross-platform if there is contention on the cache file.

embetten:
Bug fix #416 
- Move session token cache file lock to single wait handle.
- Add new session cache specific cancel message so the cancel message is not printed twice.
- Pass cancellation token to credential request handler.


cormacpayne (approved):
:shipit:

jmyersmsft (approved):

PR #488, opened 2024-03-07
Removing Publishing Steps from public build

embetten:
- Remove publishing symbols and nupkg package from public build to internal release pipeline


build/build.yml
```diff
@@ -178,20 +178,6 @@ steps:
...
-      PublishSymbols: true
-      SymbolServerType: TeamServices
-      SymbolsProduct: 'artifacts-credprovider'
-    env:
-      ARTIFACTSERVICES_SYMBOL_ACCOUNTNAME: 'microsoft'
-      ARTIFACTSERVICES_SYMBOL_USEAAD: 'false'
-
 - ${{ if eq(parameters.publish, 'true') }}:
   - task: PublishPipelineArtifact@1
     displayName: 'Upload symbols'
```
JohnSchmeichel, line 183: If the pipeline artifact has symbol files included in it would say also remove this one then?

JohnSchmeichel: Then I think all the parameters.publish usage is gone and we can remove that from all the other yamls

embetten: good point removed symbols upload and removed unused publish parameter

JohnSchmeichel:
.vsts-release.yml should be obsolete at this point? can we remove it now too?

JohnSchmeichel:
sorry one more: there is usage of PushNupkg which is obsolete across the GitHub yamls?

JohnSchmeichel (approved):

PR #485, opened 2024-03-06
Revert System.Text.Json change

embetten:
- Reverting system.text.json #393 PR due to #484.
- Adding warning message for single quotes in endpoint message.
- Did not change swix/swr dependencies.

PR #492, opened 2024-05-02
Managed Identity and Service Principal Support

embetten:
# Overview
- Added MSAL Managed Identity and Service Principal Token Providers to Microsoft.Artifacts.Authentication Library. 
- Created new endpoint `ARTIFACTS_CREDENTIALPROVIDER_FEED_ENDPOINTS` environment variable with new json schema for MI/SP required fields.
- Updated VstsBuildTaskServiceEndpointCredentialProvider to call  Microsoft.Artifacts.Authentication for MI/SP token providers.
-  Reverted #485 Changes to use system.text.json for de/serialization everywhere except for the `VSS_NUGET_EXTERNAL_FEED_ENDPOINTS` environment variable.

## Design Decisions
- Intentionally not supporting SP secrets authentication to promote security best practices. 
- The new environment variable name and json schema were created instead of reusing or extending the existing `VSS_NUGET_EXTERNAL_FEED_ENDPOINTS` to reduce password usage and clarify the environment variable will be available to our other credproviders such as the [artifacs-keyring](https://github.com/microsoft/artifacts-keyring) not just NuGet. 

## Environment Variable
 `ARTIFACTS_CREDENTIALPROVIDER_FEED_ENDPOINTS` 
```javascript
 {"endpointCredentials": [{"endpoint":"http://example.index.json", "clientId":"required", "clientCertificateSubjectName":"optional", "clientCertificateFilePath":"optional"}]}
```
- `endpoint`: required. Feed url to authenticate against. 
- `clientId`: required for both MI/SP. For user assigned managed identities enter the Entra client id. For system assigned variables set the value to `system`.  
- `clientCertificateSubjectName`: Subject Name of the certificate located in the My/ CurrentUser or LocalMachine certificate store. Optional field. Only used by SP authentication. 
- `clientCertificateFilePath`: File path location of the certificate on the machine. Optional field. Only used by SP authentication. 

Will throw error if both `clientCertificateSubjectName` or `clientCertificateFilePath` are specified.


CredentialProvider.Microsoft/Util/CertificateUtil.cs
```diff
@@ -0,0 +1,76 @@
...
+            if (string.IsNullOrWhiteSpace(subjectName))
+            {
+                logger.Info(message: Resources.InvalidCertificateInput);
+                return null;
+            }
+
+            var locations = new List<StoreLocation> { StoreLocation.LocalMachine, StoreLocation.CurrentUser };
+            foreach (var location in locations)
+            {
+                var store = new X509Store(StoreName.My, location);
```
embetten, line 21: Note this currenlty does not work on unix -  Unix LocalMachine X509Store is limited to the Root and CertificateAuthority stores.

File path works

CredentialProvider.Microsoft.Tests/CredentialProvider.Microsoft.Tests.csproj
```diff
@@ -19,4 +19,9 @@
     <ProjectReference Include="..\src\Authentication\Microsoft.Artifacts.Authentication.csproj" />
   </ItemGroup>
 
+  <PropertyGroup>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
```
JohnSchmeichel, line 23: Combine with the properties at the top

CredentialProvider.Microsoft/CredentialProviders/Vsts/IAuthUtil.cs
```diff
@@ -44,6 +46,19 @@ public AuthUtil(ILogger logger)
             this.logger = logger;
         }
 
+        public async Task<string> GetTenantIdAsync(Uri uri, CancellationToken cancellationToken)
```
JohnSchmeichel, line 49: let's discuss this one in person.. the tenant is part of the endpoint authorization which is the "official" place, and extracted from there in other places. Not opposed to the resource tenant one if that's officially supported. In either case, we really want to pull 2 values at the same time and pass them around, so should consider a struct or something to contain this info (and update other MSAL stuff to use the new TenantId stuff). Oh, and test for MSAs as well.

CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsCredentialProvider.cs
```diff
@@ -41,11 +41,15 @@ public override async Task<bool> CanProvideCredentialsAsync(Uri uri)
...
             // we should not try get credentials with this cred provider.
-            string feedEndPointsJsonEnvVar = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskExternalEndpoints);
+            string feedEndPointsJsonEnvVar = Environment.GetEnvironmentVariable(EnvUtil.EndpointCredentials);
+            string externalFeedEndPointsJsonEnvVar = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskExternalEndpoints);
             string uriPrefixesStringEnvVar = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskUriPrefixes);
             string accessTokenEnvVar = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskAccessToken);
 
-            if (string.IsNullOrWhiteSpace(feedEndPointsJsonEnvVar) == false || string.IsNullOrWhiteSpace(uriPrefixesStringEnvVar) == false || string.IsNullOrWhiteSpace(accessTokenEnvVar) == false)
+            if (string.IsNullOrWhiteSpace(feedEndPointsJsonEnvVar) == false 
+                || string.IsNullOrWhiteSpace(externalFeedEndPointsJsonEnvVar) == false
```
JohnSchmeichel, line 50: nit: please no :)

CredentialProvider.Microsoft/CredentialProviders/Vsts/IAuthUtil.cs
embetten: sounds good - I thought the auth endpoint can have non-tenant id information such as /organization

CredentialProvider.Microsoft/CredentialProviders/VstsBuildTaskServiceEndpoint/VstsBuildTaskServiceEndpointCredentialProvider.cs
```diff
@@ -61,24 +60,92 @@ public override Task<bool> CanProvideCredentialsAsync(Uri uri)
...
+                var tenantId = await AuthUtil.GetTenantIdAsync(request.Uri, cancellationToken);
+                Verbose(string.Format(Resources.UsingTenant, tenantId));
+
+                var clientCertificate = GetCertificate(matchingEndpoint);
+                Logger.Verbose(clientCertificate == null
+                    ? (Resources.ClientCertificateNotFound)
+                    : string.Format(Resources.UsingCertificate, clientCertificate.Subject));
+
+                IEnumerable<ITokenProvider> tokenProviders = await TokenProvidersFactory.GetAsync(authority);
+                tokenProviders = tokenProviders.Where(x => x.Name == "MSAL Managed Identity" || x.Name == "MSAL Service Principal").ToList();
```
JohnSchmeichel, line 96: Let's add more methods to TokenProvidersFactory where necessary: e.g. GetEndpointProvidersAsync

CredentialProvider.Microsoft/Program.cs
```diff
@@ -87,8 +87,9 @@ public static async Task<int> Main(string[] args)
                             EnvUtil.SessionTimeEnvVar,
                             EnvUtil.TokenTypeEnvVar,
                             EnvUtil.BuildTaskUriPrefixes,
-                            EnvUtil.BuildTaskAccessToken,
                             EnvUtil.BuildTaskExternalEndpoints,
+                            EnvUtil.EndpointCredentials,
```
JohnSchmeichel, line 91: Is Resources.EnvironmentVariableHelp updated to match? don't see that unless I'm blind here

CredentialProvider.Microsoft/Util/CertificateUtil.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        public static X509Certificate2 GetCertificateBySubjectName(ILogger logger, string subjectName)
+        {
+            if (string.IsNullOrWhiteSpace(subjectName))
+            {
+                logger.Info(message: Resources.InvalidCertificateInput);
+                return null;
+            }
+
+            var locations = new List<StoreLocation> { StoreLocation.LocalMachine, StoreLocation.CurrentUser };
```
JohnSchmeichel, line 18: nit: use new [] { ... } (or new list init syntax) for stuff like this.. less memory, more terse, and infers the type

CredentialProvider.Microsoft/Util/CertificateUtil.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                logger.Info(message: Resources.InvalidCertificateInput);
+                return null;
+            }
+
+            try
+            {
+                var certificate = new X509Certificate2(filePath);
+
+                if (certificate == null)
```
JohnSchmeichel, line 60: Would expect an exception if the cert can't be allocated? was this a static method before?

src/Authentication/MsalManagedIdentityTokenProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        try
+        {
+            if (string.IsNullOrWhiteSpace(tokenRequest.ClientId))
+            {
+                logger.LogTrace(string.Format(Resources.MsalClientIdError, tokenRequest.ClientId));
+                return null;
+            }
+
+            IManagedIdentityApplication mi = ManagedIdentityApplicationBuilder.Create(CreateManagedIdentityId(tokenRequest.ClientId!))
```
JohnSchmeichel, line 35: app

JohnSchmeichel, line 35: shouldn't need the ! at the end?

CredentialProvider.Microsoft/Util/FeedEndpointCredentialsParser.cs
```diff
@@ -0,0 +1,171 @@
...
+        }
+        catch (Exception e)
+        {
+            logger.Verbose(string.Format(Resources.VstsBuildTaskExternalCredentialCredentialProviderError, e));
+            return new Dictionary<string, ExternalEndpointCredentials>(StringComparer.OrdinalIgnoreCase);
+        }
+    }
+}
+
+public class ExternalEndpointCredentials
```
JohnSchmeichel, line 139: move these to the top, or a new file?

CredentialProvider.Microsoft/Util/FeedEndpointCredentialsParser.cs
```diff
@@ -0,0 +1,171 @@
...
+        {
+            logger.Verbose(string.Format(Resources.VstsBuildTaskExternalCredentialCredentialProviderError, e));
+            return new Dictionary<string, EndpointCredentials>(StringComparer.OrdinalIgnoreCase); ;
+        }
+    }
+
+
+    public static Dictionary<string, ExternalEndpointCredentials> ParseExternalFeedEndpointsJsonToDictionary(ILogger logger)
+    {
+        string feedEndPointsJson = Environment.GetEnvironmentVariable(EnvUtil.BuildTaskExternalEndpoints);
```
JohnSchmeichel, line 82: endpoint not endPoints... endpoint is 1 word, not 2 (though end and point are also words, but that's irrelevant)

src/Authentication/MsalServicePrincipalTokenProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+                if (!CanGetToken(tokenRequest))
+                {
+                    logger.LogTrace("InvalidInputs");
+                    return null;
+                }
+
+                var app = ConfidentialClientApplicationBuilder
+                    .Create(tokenRequest.ClientId)
+                    .WithHttpClientFactory(appConfig.HttpClientFactory)
+                    .WithCertificate(tokenRequest.ClientCertificate)
```
JohnSchmeichel, line 40: we need to set the "sendX5c" or whatever

src/Authentication/MsalManagedIdentityTokenProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        this.appConfig = appConfig;
+        this.logger = logger;
+    }
+
+    public string Name => "MSAL Managed Identity";
+
+    public bool IsInteractive => false;
+
+    public bool CanGetToken(TokenRequest tokenRequest) =>
```
JohnSchmeichel, line 22: we should check against system here, so if the env var is not set or not system we don't try, otherwise we'll try

CredentialProvider.Microsoft/Util/FeedEndpointCredentialsParser.cs
```diff
@@ -0,0 +1,171 @@
...
+                var urlEncodedEndpoint = endpointUri.AbsoluteUri;
+                if (!credsResult.ContainsKey(urlEncodedEndpoint))
+                {
+                    credsResult.Add(urlEncodedEndpoint, credentials);
+                }
+            }
+
+            return credsResult;
+        }
+        catch (Exception e)
```
JohnSchmeichel, line 72: ex

CredentialProvider.Microsoft/Util/FeedEndpointCredentialsParser.cs
```diff
@@ -0,0 +1,171 @@
...
```
JohnSchmeichel, line 131: ex

CredentialProvider.Microsoft/CredentialProviders/VstsBuildTaskServiceEndpoint/VstsBuildTaskServiceEndpointCredentialProvider.cs
JohnSchmeichel: discussed in person, a new factory for the endpoint case to just create the providers needed and avoid the filtering out

CredentialProvider.Microsoft/Util/FeedEndpointCredentialsParser.cs
embetten: done

CredentialProvider.Microsoft/Util/FeedEndpointCredentialsParser.cs
embetten: done

CredentialProvider.Microsoft/Util/FeedEndpointCredentialsParser.cs
embetten: top

src/Authentication/MsalManagedIdentityTokenProvider.cs
embetten: not needed for the compiler, removed

CredentialProvider.Microsoft/Util/CertificateUtil.cs
embetten: okay!

CredentialProvider.Microsoft/Program.cs
embetten: updated

CredentialProvider.Microsoft/CredentialProviders/VstsBuildTaskServiceEndpoint/VstsBuildTaskServiceEndpointCredentialProvider.cs
embetten: done

CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsCredentialProvider.cs
embetten: fixed :)

CredentialProvider.Microsoft/CredentialProviders/Vsts/IAuthUtil.cs
embetten: moved to struct

CredentialProvider.Microsoft/Util/FeedEndpointCredentialsParser.cs
embetten: done!

CredentialProvider.Microsoft/CredentialProviders/Vsts/IAuthUtil.cs
embetten: resolving - reopen or let me know if you have any feedback on the implementation

CredentialProvider.Microsoft/Util/FeedEndpointCredentialsParser.cs
```diff
@@ -0,0 +1,171 @@
...
+                    break;
+                }
+
+                if (credentials.ClientId == null)
+                {
+                    logger.Verbose(Resources.EndpointParseFailure);
+                    break;
+                }
+
+                if(credentials.CertificateSubjectName != null && credentials.CertificateFilePath != null)
```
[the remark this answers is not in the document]
JohnSchmeichel: lol, yeah, this one is the missing space between if(

CredentialProvider.Microsoft/Util/CertificateUtil.cs
embetten: why throw vs return null or empty in the util layer? Regardless, if certificate information is provided and we fail to get a certificate, we should definitely have some user-friendly message about why - i have updated the build task message to a info instead of verbose to make it clearer.

src/Authentication/MsalServicePrincipalTokenProvider.cs
embetten: done

CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsBuildTaskMsalTokenProvidersFactory.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Artifacts.Authentication;
+using Microsoft.Extensions.Logging;
+using NuGetCredentialProvider.Util;
+
+namespace NuGetCredentialProvider.CredentialProviders.Vsts;
+
+internal class VstsBuildTaskMsalTokenProvidersFactory : ITokenProvidersFactory
```
JohnSchmeichel, line 11: Move this class under the VstsBuildTaskServiceEndpoint folder where it's used

embetten: yes

CredentialProvider.Microsoft/Program.cs
```diff
@@ -87,8 +88,9 @@ public static async Task<int> Main(string[] args)
...
```
JohnSchmeichel, line 92: Order seems off from help text...

CredentialProvider.Microsoft/CredentialProviders/VstsBuildTaskServiceEndpoint/VstsBuildTaskMsalTokenProvidersFactory.cs
```diff
@@ -0,0 +1,38 @@
...
+                {
+                    // We ignore containsPii param because we are passing in enablePiiLogging below.
+                    logger.LogTrace("MSAL Log ({level}): {message}", level, message);
+                },
+                enablePiiLogging: EnvUtil.GetLogPIIEnabled()
+            )
+            .Build();
+
+        return Task.FromResult(MsalTokenProviders.Get(app, logger)
+            .Where(x => x.Name == "MSAL Managed Identity" || x.Name == "MSAL Service Principal"));
```
JohnSchmeichel, line 36: Just construct the two we need here instead of filtering... think this means we can avoid creating the app entirely above?

JohnSchmeichel:
> using System;

add license header

---
Refers to: CredentialProvider.Microsoft/CredentialProviders/VstsBuildTaskServiceEndpoint/VstsBuildTaskMsalTokenProvidersFactory.cs:1 in 0dceaff. [](commit_id = 0dceaff4b5011ec085c15f3ae33bd0c967a108f0, deletion_comment = False)

JohnSchmeichel:
> using System;

same for all new (non test?) files

---
In reply to: [2116411091](https://github.com/microsoft/artifacts-credprovider/pull/492#issuecomment-2116411091) [](http://example.com/codeflow?ancestors=2116411091)

---
Refers to: CredentialProvider.Microsoft/CredentialProviders/VstsBuildTaskServiceEndpoint/VstsBuildTaskMsalTokenProvidersFactory.cs:1 in 0dceaff. [](commit_id = 0dceaff4b5011ec085c15f3ae33bd0c967a108f0, deletion_comment = False)

src/Authentication/ITokenProvidersFactory.cs
```diff
@@ -6,5 +6,5 @@ namespace Microsoft.Artifacts.Authentication;
 
 public interface ITokenProvidersFactory
 {
-    Task<IEnumerable<ITokenProvider>> GetAsync(Uri authority);
+    Task<IEnumerable<ITokenProvider>> Get(Uri authority);
```
JohnSchmeichel, line 9: >Get [](http://example.com/codeflow?start=38&length=3)

does this need to change? will be breaking if others update their package version

JohnSchmeichel: would avoid any of these unless required for the public types

src/Authentication/MsalServicePrincipalTokenProvider.cs
```diff
@@ -0,0 +1,57 @@
+﻿using Microsoft.Extensions.Logging;
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Artifacts.Authentication
+{
+    public class MsalServicePrincipalTokenProvider : ITokenProvider
```
JohnSchmeichel, line 6: >MsalServicePrincipalTokenProvider [](http://example.com/codeflow?start=17&length=33)

consider updating MsalAuthenticationTests as well if it makes sense.. the idea there is it does a real token acquisition, so the tests are ignored by default, but can be run on-demand locally to ensure the token providers work

JohnSchmeichel (approved):
:shipit:

CredentialProvider.Microsoft/CredentialProviders/VstsBuildTaskServiceEndpoint/VstsBuildTaskMsalTokenProvidersFactory.cs
embetten: I would still need to create the app above since the two we need to construct take in the app as a constructor parameter. Additionally, the ITokenProviderFactory interface is not super friendly for the proposed scenario since it is returning a Task<IEnumerable<ITokenProvider>>

JohnSchmeichel: cache should be added which should help a bit since the async was required for that.. then it's just returning an IEnumerable

JohnSchmeichel:
please address the build warnings as well, there are a few about package conflicts. surprised these are not failing the build :(

CredentialProvider.Microsoft/Program.cs
JohnSchmeichel: e.g.

```

Build Provider Access Token
    VSS_NUGET_EXTERNAL_FEED_ENDPOINTS
        The Personal Access Token that will be returned as credentials by
        the build provider.

Build Provider Service Endpoint Json
    ARTIFACTS_CREDENTIALPROVIDER_FEED_ENDPOINTS
        Json that contains an array of service endpoints, usernames and
        access tokens to authenticate endpoints in nuget.config.
        Example: {"endpointCredentials": [{"endpoint":"http://example.index.json",
        "username":"optional", "password":"accesstoken"}]}

Artifacts Service Endpoint Json
    VSS_NUGET_ACCESSTOKEN
        Json that contains an array of service endpoints, client ids, and certificate
        information to authenticate endpoints for Azure Managed Identities
        and Service Principals.
        Example: {"endpointCredentials": [{"endpoint":"http://example.index.json",
        "clientId":"required", "clientCertificateFilePath":"optional","
        clientCertificateSubjectName": "optional" }]}

```

src/Authentication/MsalServicePrincipalTokenProvider.cs
embetten: I added tests and allowed client id and certificate to be passed with an env variable, though additional work to add a MI/SP to the feed

CredentialProvider.Microsoft/CredentialProviders/VstsBuildTaskServiceEndpoint/VstsBuildTaskMsalTokenProvidersFactory.cs
embetten: I can add it back but I initially removed it due to: https://learn.microsoft.com/en-us/entra/msal/java/advanced/managed-identity#caching not being supported

JohnSchmeichel: but service principals it should work?

JohnSchmeichel: actually, the cache needs to be registered on the app which you're creating inside the token providers now. so either no cache, or create the apps on the outside and pass them in

CredentialProvider.Microsoft/Program.cs
embetten: yup! fixed

CredentialProvider.Microsoft/CredentialProviders/VstsBuildTaskServiceEndpoint/VstsBuildTaskMsalTokenProvidersFactory.cs
embetten: removed - I think this is a much larger change in the [vstsbuildendpoint cred as well](https://github.com/microsoft/artifacts-credprovider/blob/74fe273afe1270081d69cc219133afd0d230fba6/CredentialProvider.Microsoft/CredentialProviders/VstsBuildTaskServiceEndpoint/VstsBuildTaskServiceEndpointCredentialProvider.cs#L48). to support caching for only service principal returned tokens.

Issue #464, opened 2024-01-10
The messages logged by the credential providers are not localized.


kartheekp-ms:
### Discussed in https://github.com/microsoft/artifacts-credprovider/discussions/433

<div type='discussions-op-text'>

<sup>Originally posted by **kartheekp-ms** July  6, 2023</sup>
When I execute dotnet restore command and NuGet is trying to access a private feed, credential provider displays following logs to the console.

```
log  :    [CredentialProvider]ATTENTION: User interaction required.

log  :

log  :    **********************************************************************

log  :

log  :    To sign in, use a web browser to open the page https://microsoft.com/devicelogin and enter the code {REMOVED} to authenticate.
```
Looking at the repository, there is no code related to localization, but just wanted to confirm that currently log messages are not localized. Are there any plans to enable localization in the future?</div>

kartheekp-ms:
As you can see in the example below, I changed the language code in the .NET CLI to 'es'. While the NuGet-related log messages respected the language setting, the credential provider logs were still printed in English.

![image](https://github.com/microsoft/artifacts-credprovider/assets/52756182/7c7dd573-665e-4559-bf3e-0155c18bafd9)

embetten:
reopening - not stale

PR #500, opened 2024-07-08
Broker: Add ability to pass parent window handle explicitly for Windows broker auth

seclerp:
We have a headless process that hosts the NuGet credential plugin manager and we are facing issues with the default broker logic which tries to obtain the current process' window handle, as we need to pass a native window handle of another (non-.NET) desktop app process. I think the change will be quite beneficial in such a scenario.

I'm open to discussion in case env-based handle propagation is not the best approach.

### DIFF
--- a/CredentialProvider.Microsoft/Util/EnvUtil.cs
+++ b/CredentialProvider.Microsoft/Util/EnvUtil.cs
@@ -35,7 +35,7 @@ namespace NuGetCredentialProvider.Util
         public const string MsalFileCacheEnvVar = "NUGET_CREDENTIALPROVIDER_MSAL_FILECACHE_ENABLED";
         public const string MsalFileCacheLocationEnvVar = "NUGET_CREDENTIALPROVIDER_MSAL_FILECACHE_LOCATION";
         public const string MsalAllowBrokerEnvVar = "NUGET_CREDENTIALPROVIDER_MSAL_ALLOW_BROKER";
-        public const string MsalBrokerWindowEnvVar = "NUGET_CREDENTIALPROVIDER_MSAL_BROKER_WINDOW";
+        public const string MsalBrokerWindowEnvVar = "ARTIFACTS_CREDENTIALPROVIDER_MSAL_BROKER_WINDOW";
 
         public const string EndpointCredentials = "ARTIFACTS_CREDENTIALPROVIDER_FEED_ENDPOINTS";
         public const string BuildTaskExternalEndpoints = "VSS_NUGET_EXTERNAL_FEED_ENDPOINTS";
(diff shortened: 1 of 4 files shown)


seclerp:
@microsoft-github-policy-service agree company="JetBrains"

CredentialProvider.Microsoft/Util/EnvUtil.cs
```diff
@@ -35,6 +35,7 @@ public static class EnvUtil
         public const string MsalFileCacheEnvVar = "NUGET_CREDENTIALPROVIDER_MSAL_FILECACHE_ENABLED";
         public const string MsalFileCacheLocationEnvVar = "NUGET_CREDENTIALPROVIDER_MSAL_FILECACHE_LOCATION";
         public const string MsalAllowBrokerEnvVar = "NUGET_CREDENTIALPROVIDER_MSAL_ALLOW_BROKER";
+        public const string MsalBrokerWindowEnvVar = "NUGET_CREDENTIALPROVIDER_MSAL_BROKER_WINDOW";
```
embetten, line 38: nit: can we rename this env varialble to ARTIFACTS_CREDENTIALPROVIDER - this way it will be clear its available to our other protocol credprovider wrappers such as the python https://github.com/microsoft/artifacts-keyring

embetten (changes requested):

seclerp:
I have a question - what is a proper workflow for internationalization? Should I update the default `Resources.resx` only?

seclerp:
Hi @embetten, do you need something else from my side, any concerns? Is there something that prevents this stuff from being merged?

embetten:
> I have a question - what is a proper workflow for internationalization? Should I update the default `Resources.resx` only?

Currently this is not localized. Any strings to be logged should be added there so they will be localized in the future ( #464 )

> Hi @embetten, do you need something else from my side, any concerns? Is there something that prevents this stuff from being merged?

Conflict resolutions would be appreciated. The only blocker was our internal validation.

src/Authentication/AzureArtifacts.cs
```diff
@@ -53,6 +53,11 @@ public static PublicClientApplicationBuilder WithBroker(this PublicClientApplica
                 })
             .WithParentActivityOrWindow(() => GetConsoleOrTerminalWindow());
```
embetten, line 54: does this need to be something like:
```
.WithParentActivityOrWindow(() =>  parentWindowHandle ?? GetConsoleOrTerminalWindow());
```
?

seclerp: You're right. Seems like I've missed that stuff during conflict resolution

embetten (approved):
Thank you!

seclerp:
Hi folks, it's been a while, can we merge it? :)

embetten:
@seclerp - apologies for the delay here. Due to some unforeseen infrastructure blockers on our end, we were unable to merge your PR at the time. If you are able to recreate or resubmit the PR, we would appreciate this contribution. Thanks!

PR #502, opened 2024-07-24
Add parameterless constructor for Microsoft.Artifacts.Authentication.TokenRequest 

NoDataAvailable:
The `Uri` property in TokenRequest is unused. Remove it to make it less confusing.


embetten (approved):
LGTM

PR #504, opened 2024-08-01
Don't rebuild during pack

JohnSchmeichel:
The build occurs in many phases and was relying on incremental builds not re-building outputs, which is not deterministic. Properties were not consistently passed at each step which results in outputs being flagged as out-of-date, causing rebuilds and loosing signing. These properties are now consistently passed, and --no-build used to ensure signed files are not overridden.


embetten (approved):
thank you

PR #522, opened 2024-10-09
Use MSTest meta package

stan-sz:
Follow https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-mstest-getting-started to use the MSTest meta package that includes other MSTest packages, including analyzers.


stan-sz:
See also https://www.youtube.com/watch?v=9Jz47Ze9LOI

embetten (approved):

PR #528, opened 2024-10-15
Don't attempt to install .NET 8 without explicit variable

cobya:
With v1.3.0 not being fully released yet, attempting download of Net8 bits may fail if not explicitly setting v1.3.0-alpha in `installcredprovider.sh`. Update the variable check to avoid "default" downloads of .NET 8.


embetten (approved):
Approve

PR #550, opened 2025-03-20
Remove mention of example

mjsabby:
This poorly placed comment is causing people to download Windows binaries on macOS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,10 +61,6 @@ Using the above is recommended, but as per [NuGet's plugin discovery rules](http
 
 [Linux or Mac helper script](helpers/installcredprovider.sh)
 
-Examples:
-- `wget -qO- https://aka.ms/install-artifacts-credprovider.sh | bash`
-- `sh -c "$(curl -fsSL https://aka.ms/install-artifacts-credprovider.sh)"`
-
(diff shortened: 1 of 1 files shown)


prezaei:
Looks to me what is missing is something like this:

```
USE_NET6_ARTIFACTS_CREDENTIAL_PROVIDER=false
USE_NET8_ARTIFACTS_CREDENTIAL_PROVIDER=true
ARTIFACTS_CREDENTIAL_PROVIDER_RID=osx-arm64
```

embetten:
@mjsabby the examples download the current default platform agnostic, non-self-contained version of the credprovider that requires the dotnet runtime. The self-contained platform specific binaries were just released in the latest version this week. We are still in an opt-in phase while we confirm the experience. While we work to make the self-contained and the .NET 8 experience the default, please use the variables mentioned by @prezaei.

> Looks to me what is missing is something like this:
> 
> ```
> export USE_NET6_ARTIFACTS_CREDENTIAL_PROVIDER=false
> export USE_NET8_ARTIFACTS_CREDENTIAL_PROVIDER=true
> export ARTIFACTS_CREDENTIAL_PROVIDER_RID=osx-arm64
> ```

As I am planning to update the README with more detailed self-contained examples for both windows or non windows

PR #558, opened 2025-05-02
[Draft] Major Release v2.0.0rc0

cobya:
See #562 for a list of upcoming changes.
- Use self-contained binaries by default when installing. Existing environment variables can be used to override this behavior.
- Versioned install scripts for each release version. New installs will use a helper fetch script to get the correct install script version for full backwards compatability.
- Remove end of support .NET Core 3.1 and .NET Framework 4.6.1
- Add some SHA256 validation to downloads


Build.props
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <CredentialProviderVersion>1.4.1</CredentialProviderVersion>
+    <CredentialProviderVersion>2.0.0</CredentialProviderVersion>
     <VersionSuffix></VersionSuffix>
-    <TargetFrameworks>netcoreapp3.1;net461;net481;net6.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net481;net6.0;net8.0</TargetFrameworks>
   </PropertyGroup>
   <!-- Remove with netcoreapp3.1; required for self-contained build -->
```
Frulfump, line 8: Looks like this should be removed as well?

build/build.yml
Frulfump, line 1: There are some additional .NET Core 3.1 things left in this file around line 140 and down

Build.props
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <CredentialProviderVersion>1.4.1</CredentialProviderVersion>
+    <CredentialProviderVersion>2.0.0</CredentialProviderVersion>
     <VersionSuffix></VersionSuffix>
-    <TargetFrameworks>netcoreapp3.1;net461;net481;net6.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net481;net6.0;net8.0</TargetFrameworks>
```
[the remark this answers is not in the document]
cobya: It's on our radar to deprecate .NET 6 as well, however we've previously announced in #523 and #562 that 2.0.0 would remove 3.1 and 4.6.1, and for a breaking change such as this when users are automatically opted into the new version we would want to give as much advance notice as possible for this. With the move to self-contained binaries by default, specific .NET versions will become less important and can be removed more easily.

build/build.yml
cobya: Wasn't quite ready for reviews when you got this, removed 👍

Build.props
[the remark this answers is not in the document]
Frulfump: Ah that sounds fine. Support for .NET 6 is being removed from some places like https://github.com/actions/runner-images/issues/12241 in August. Thanks for the additional info!

build/build.yml
```diff
@@ -143,74 +137,8 @@ steps:
...
-
+# Clean target folder and create netfx releases
 # NOTE: Changing the name of the zip will break things. Please don't do it.
-- task: ArchiveFiles@2
-  displayName: Create zip
-  inputs:
-    rootFolderOrFile: '$(Build.ArtifactStagingDirectory)\tarball\'
-    includeRootFolder: false
-    archiveType: 'zip'
-    archiveFile: '$(Build.ArtifactStagingDirectory)\$(BuildConfiguration)\Microsoft.NuGet.CredentialProvider.zip'
```
embetten, line 212: I think not producing a Microsoft.NuGet.CredentialProvider.zip/tar assets will break things - we might need to rename net 6 or net 8 to Microsoft.NuGet.CredentialProvider.zip/tar

cobya: The NuGet plugin should rely on .NET Framework only correct? We're still publishing `Microsoft.NuGet.NetFx48.CredentialProvider.zip`

PR #566, opened 2025-06-18
 Add configuration to publish cred provider as tool.

morended:
This PR introduces packaging credprovider as dotnet tool.


Build.props
```diff
@@ -3,7 +3,14 @@
   <PropertyGroup>
     <CredentialProviderVersion>1.4.1</CredentialProviderVersion>
     <VersionSuffix></VersionSuffix>
-    <TargetFrameworks>netcoreapp3.1;net461;net481;net6.0;net8.0</TargetFrameworks>
+    <!--PackAsTool is set to true for net8.0 framework-->
+    <PackAsTool>false</PackAsTool>
```
embetten, line 7: Can we try moving the package id and other nuspec infromation to a separate nuspec file like we do for https://github.com/microsoft/artifacts-credprovider/blob/master/CredentialProvider.Microsoft/CredentialProvider.Microsoft.nuspec 

We will need all the copywrite and compliance information in the nuspec for OSS compliance purposes

Build.props
```diff
@@ -3,7 +3,14 @@
...
+    <ToolCommandName>nuget-plugin-microsoft-artifacts-credential-provider</ToolCommandName>
+    <PackageOutputPath>./nupkg</PackageOutputPath>
```
embetten, line 9: Can we split the toll command specific information to a seperate property group like we do in line 16 below

Build.props
morended: Moved PackAsTool config to nuspec

Build.props
morended: By default "PackAsTool" is set to false. The existing behavior to generate and publish non tool binaries remains same. To pack credprovider as tool, we have to run `dotnet pack -p:PackAsTool=true`. This will create dotnet tool wrapped in a nupkg.

CredentialProvider.Microsoft/CredentialProvider.Tool.Microsoft.nuspec
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
```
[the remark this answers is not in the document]
morended: Renamed file.

PR #573, opened 2025-07-14
Fix unwanted warning when installing using shell script

ronlv4:
The test and the log are inconsistent in logic, tests if the variable is empty and then log that it is set.

fixed to log only if the variable is not empty.

### DIFF
--- a/helpers/installcredprovider.sh
+++ b/helpers/installcredprovider.sh
@@ -25,7 +25,7 @@ if [ ! -z ${ARTIFACTS_CREDENTIAL_PROVIDER_RID} ]; then
     ;;
   esac
 
-  if [ -z ${USE_NET6_ARTIFACTS_CREDENTIAL_PROVIDER} ]; then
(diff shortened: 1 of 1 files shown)


helpers/installcredprovider.sh
```diff
@@ -99,4 +99,4 @@ if ! curl -H "Accept: application/octet-stream" \
         exit 1
 fi
 
-echo "INFO: credential provider netcore plugin extracted to $HOME/.nuget/"
\ No newline at end of file
+echo "INFO: credential provider netcore plugin extracted to $HOME/.nuget/"
```
Frulfump, line 102: I think this diff means something changed with end of file. There's no `.editorconfig` in this repo sadly opened new issue #574 to track that it should contain among other thing `insert_final_newline = true`

ronlv4: you were this was a new line at the end of the file added automatically, maybe by VIM I'm not sure.
for now removed it since it's out of scope for this PR.

PR #597, opened 2025-11-07
Service principal authentication with client secret

LapNik:
Adds the option to specify a client secret in the endpoint credential structures read from the ARTIFACTS_CREDENTIALPROVIDER_FEED_ENDPOINTS environment variable.

A client certificate may be difficult to use in some CICD scenarios, so this adds another option for service principal authentication. Users should still prefer certificates over secrets.


LapNik:
@microsoft-github-policy-service agree

PR #606, opened 2025-11-20
Move broker auth to opt out releases/2.0.0

embetten:
move broker auth to default or "opt out" after internal validations  
#605


pwoosam (approved):
Not sure why the build is failing, but it's unrelated to these changes.

PR #636, opened 2026-02-05
scripts: enable boolean parameters to PowerShell release scripts 

embetten:
- adds a script to test install script release
- fixes breaking change to allow scripts to pass colon syntax 
- I also wanted to propose this test-*.ps1 script as a future approach to replace our current install script validation. This way we can run the tests easier for both local and ci dev


embetten:
closing - team decided against custom parsing in order to support : format.

PR #668, opened 2026-04-21
[DO NOT MERGE] [Experimental] Fix macOS broker hanging on non-Intune joined devices

embetten:
Upstream issue: AzureAD/microsoft-authentication-library-for-dotnet#5940

Problem
On macOS devices without a Microsoft SSO extension (e.g. Company Portal), calling `AcquireTokenInteractive` through the broker causes an unrecoverable hang. The native MSAL broker runtime triggers FallbackToNativeMsal when no SSO extension handles the request, which appears to perform a synchronous GCD dispatch (dispatch_sync) back to the main queue - deadlocking because the main thread is already blocked waiting for the broker result. This affects any macOS user who has broker enabled but does not have Company Portal or another Microsoft SSO extension installed.

### Workarounds
All macOs non intune joined users must set `ARTIFACTS_CREDENTIALPROVIDER_MSAL_ALLOW_BROKER=false`

### PR Proposed Fixes

| Approach | Tradeoff |
|---|---|
| **SSO extension detection** — skip broker if `pluginkit` finds no Microsoft SSO extension | Brittle; depends on Apple CLI tool and knowledge of MSAL internals |
| **Timeout guard** — abandon broker via `Task.WhenAny` if it doesn't respond | Main thread stays stuck; adds up to 90s latency before fallback |

### Preferred 
- Native broker / integration fix 
- MSAL exposes a `BrokerOptions` setting to disable native fallback
- MSAL throws `MsalException` instead of deadlocking when no SSO extension is found?

### DIFF
--- a/src/Authentication.Tests/TokenProviderTests.cs
+++ b/src/Authentication.Tests/TokenProviderTests.cs
@@ -175,4 +175,139 @@ public class TokenProviderTests
         tokenRequest.ClientId = null;
         Assert.IsFalse(tokenProvider.CanGetToken(tokenRequest));
     }
+
+    [TestMethod]
+    public async Task MsalBrokerInteractive_ReturnsNull_OnMsalServiceException()
+    {
+        var app = CreateRealApp();
+        // IsUserInteractive() should pass on real PCA, and AcquireTokenInteractive
+        // will throw MsalServiceException during ExecuteAsync.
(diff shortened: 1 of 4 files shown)


embetten:
Closing - MSAL will provide the fix. We will utilize the env workaround until that is available

PR #672, opened 2026-04-27
Add SPS endpoint allowlist and secure probe client

embetten:
This PR tightens how the credential provider handles discovery and token exchange as part of ongoing Entra identity hardening:

- Credentialless probe client: Introduced [HttpClientFactory.Probe](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/10c8e557c8/resources/app/out/vs/code/electron-browser/workbench/workbench.html), a dedicated HttpClient for discovery requests that does not set UseDefaultCredentials. This ensures Windows Integrated Authentication credentials are not sent when probing unknown endpoints for Azure DevOps headers.

- HTTPS-only probing: Unknown hosts are now only probed over HTTPS. HTTP endpoints that aren't in the known hosts list are immediately treated as external without making a network call.

- SPS endpoint allowlist: Added validation of the X-VSS-AuthorizationEndpoint header value before exchanging bearer tokens. The endpoint must match a known Azure DevOps SPS hostname (e.g., *.vssps.visualstudio.com, vssps.dev.azure.com). This prevents session token exchange against unexpected endpoints.

Testing:

258 unit tests passing (including 19 new tests for SPS allowlist validation)
Integration tested against a live Azure DevOps feed across all three target frameworks (net481, net6.0, net8.0) with broker, no-broker, and Entra token opt-in configurations


CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsSessionTokenClient.cs
```diff
@@ -18,6 +19,19 @@ public class VstsSessionTokenClient : IVstsSessionTokenClient
...
         private const string TokenScope = "vso.packaging_write vso.drop_write";
 
+        // Known Azure DevOps SPS hostnames that are trusted to receive bearer tokens.
+        // This prevents token exfiltration via a malicious X-VSS-AuthorizationEndpoint header.
+        public static readonly string[] AllowedSpsHosts = new[]
+        {
+            "app.vssps.visualstudio.com",          // Azure DevOps production
+            ".vssps.visualstudio.com",             // Azure DevOps production (suffix)
+            "vssps.dev.azure.com",                 // Azure DevOps production
+            "app.vssps.dev.azure.com",             // Azure DevOps production
```
embetten: we need to validate this list of hosts

embetten: I updated the list with production hosts, we just need to confirm if there are any more ppe hosts

cobya (approved):

CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsSessionTokenClient.cs
```diff
@@ -81,10 +90,12 @@ public async Task<string> CreateSessionTokenAsync(VstsTokenType tokenType, DateT
 
             if (!IsAllowedSpsEndpoint(spsEndpoint))
             {
-                logger.Log(NuGet.Common.LogLevel.Error, true,
+                var message =
                     $"SPS authorization endpoint '{spsEndpoint}' is not a known Azure DevOps host. " +
-                    "Aborting session token exchange to prevent bearer token exfiltration.");
-                return null;
+                    "Aborting session token exchange to prevent bearer token exfiltration.";
```
cobya: I'm not convinced we should explicitly say this 😅 I'll leave it to you though

cobya (approved):

CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsSessionTokenClient.cs
embetten: removed !

cobya (approved):

Issue #664, opened 2026-04-16
[Bug]: Silent WAM entra token acquisition doesn't work


amaher-ms:
### Description

MsalSilentTokenProvider always receives the non-broker IPublicClientApplication, so it can never acquire tokens
silently from the WAM broker cache. This means every invocation requires interactive authentication when the broker is enabled and `ARTIFACTS_CREDENTIALPROVIDER_RETURN_ENTRA_TOKENS` is set (session cache not used).

In MsalTokenProvidersFactory.cs, two separate IPublicClientApplication instances are created:

 * app - without broker
 * appInteractiveBroker - with broker

In MsalTokenProviders.cs, the silent provider receives `app`:

```
yield return new MsalSilentTokenProvider(app, logger);              // non-broker
yield return new MsalBrokerInteractiveTokenProvider(appInteractiveBroker, logger); // broker
```

Because `app` has IsBrokerEnabled = false, the #if INCLUDE_BROKER block in MsalSilentTokenProvider that adds 
OperatingSystemAccount never activates, and GetAccountsAsync() never queries WAM. The MSAL file cache is also empty because broker interactive auth stores tokens in WAM's cache, not the file cache.

### Environment (OS/platform or image)

Windows 11

### Tool versions

credprovider 2.0.1

### Client/protocol path in use

None

embetten:
This fix was published as pre-release https://github.com/microsoft/artifacts-credprovider/releases/tag/v2.0.2 please try it out and let me know if this resolves the issue!

embetten:
closing - released and switched to latest please reopen if you run into any further issues. We will have a follow up version to address the WSL issue addressed in 676

PR #675, opened 2026-05-05
Fix silent auth failure in WSL caused by broker token provider

embetten:
WSL reports as Linux but has no msalruntime daemon, causing MsalSilentTokenProvider to fail when using appInteractiveBroker.

- Add PlatformInformation.IsWSL() to detect WSL via RuntimeInformation.IsOSPlatform(Linux) and WSL_DISTRO_NAME env var
- Use non-broker app for silent auth on WSL, matching behavior before 0cddac6
- Add tests for WSL and non-WSL silent auth provider selection

### Problem
Commit #664 (#666) changed `MsalSilentTokenProvidert`o use `appInteractiveBroker ?? app` so that the broker's token cache is queried on silent auth. This works correctly on Windows and macOS, but breaks Windows Subsystem for Linux (WSL) due to stdout corruption from MSAL linux broker subprocesses. 
Related MSAL issue: AzureAD/microsoft-authentication-library-for-dotnet#5979

### work around until MSAL responds
Add `PlatformInformation.IsWSL()` which returns true when the runtime OS is Linux and the WSL_DISTRO_NAME environment variable is set (always present in WSL sessions, never set on native Linux/macOS/Windows).
In `MsalTokenProviders.Get()` use the non-broker app for silent auth when running in WSL, restoring the pre-#664 behavior for WSL users while keeping broker-accelerated silent auth on Windows and macOS.

### Testing
Added unit tests covering silent provider selection under WSL and non-WSL conditions with both empty and warm MSAL caches.


MicrosoftCredentialProvider.sln
justinmsievers, line 2: is this all just line endings?

justinmsievers (approved):

MicrosoftCredentialProvider.sln
embetten: yes

PR #711, opened 2026-08-13
Preserve executable permissions for macOS installs

### LINKED ISSUE #709: [Bug]: Install script does not respect +x on MacOs
### Description

mac install script does not maintain +x bit on the executable during decompression. See original report and issue https://github.com/microsoft/artifacts-credprovider/issues/663#issuecomment-5094292639. 

looks like tar on the .zip file is the issue in the install script logic. Known workaround - manual installation.

### Environment (OS/platform or image)

MacOs-*

### Tool versions

install scripts.

### Client/protocol path in use

NuGet plugin path (nuget/plugins)

embetten:
## Summary

- Use `unzip` with a secure temporary file for macOS archives to preserve the credential provider apphost's executable bit.
- Retain the existing `tar` extraction path as a fallback.
- Validate executable permissions in bash and POSIX shell installation tests.
- Document installer requirements and musl/Alpine behavior.

Fixes #709

## Validation

- Updated direct install-script validation for bash and POSIX shell paths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,14 @@ Examples:
 - `wget -qO- https://aka.ms/install-artifacts-credprovider.sh | bash`
 - `sh -c "$(curl -fsSL https://aka.ms/install-artifacts-credprovider.sh)"`
 
+Script requirements:
+- A POSIX-compatible shell, `curl` with HTTPS certificates, and a gzip-capable `tar`.
+- Linux uses `ldd` and `grep` to detect musl. If either is unavailable, set `ARTIFACTS_CREDENTIAL_PROVIDER_NON_SC=true`.
+- macOS should have `mktemp` and `unzip` to preserve the self-contained apphost's executable bit; otherwise, it may require `chmod +x`.
+- Alpine and other musl systems require the .NET 8 runtime because the script selects the runtime-dependent archive. Do not force a `linux-x64` or `linux-arm64` RID; those assets require glibc.
+
+Microsoft-hosted Ubuntu and macOS images include these tools. Minimal Ubuntu or Alpine containers may need `ca-certificates` and `curl` installed first.
+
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -102,6 +102,14 @@ Examples:
 - `wget -qO- https://aka.ms/install-artifacts-credprovider.sh | bash`
 - `sh -c "$(curl -fsSL https://aka.ms/install-artifacts-credprovider.sh)"`
 
+Script requirements:
+- A POSIX-compatible shell, `curl` with HTTPS certificates, and a gzip-capable `tar`.
+- Linux uses `ldd` and `grep` to detect musl. If either is unavailable, set `ARTIFACTS_CREDENTIAL_PROVIDER_NON_SC=true`.
```
cobya: nit: might want to note that this will require .NET

README.md
```diff
@@ -102,6 +102,14 @@ Examples:
...
+- macOS should have `mktemp` and `unzip` to preserve the self-contained apphost's executable bit; otherwise, it may require `chmod +x`.
```
jmyersmsft: Would it not be simpler to have the script unconditionally `chmod +x` the executable, and not special-case extraction?

helpers/installcredprovider.sh
```diff
@@ -174,15 +174,48 @@ if [ -d "${NUGET_PLUGIN_DIR}/netcore/CredentialProvider.Microsoft" ]; then
...
+}
+
+extracted=""
+# macOS (osx-*) assets ship as .zip. Prefer unzip for these because, unlike tar, it
+# preserves the executable (+x) bit stored in the archive. This branch only affects .zip
+# assets, treats unzip as optional (guarded by command -v), and falls back to the original
+# tar behavior if unzip is unavailable or fails for any reason.
+case "$FILE" in
+*.zip)
+  if command -v unzip >/dev/null 2>&1; then
```
cobya: do we want a warning here if unzip doesn't exist since it also falls back?

PR #715, opened 2026-09-22
Harden credential endpoint validation

embetten:
## Summary
Strengthens validation around endpoints involved in credential acquisition and return flows.

## Changes
- Restricts credential handling to trusted or explicitly configured feed hosts in the standard VSS credential provider; build-task credential routing remains unchanged.
- Validates Azure DevOps service endpoints used during credential flows.
- Applies validation consistently across Entra and session-token flows.
- Preserves supported custom-host scenarios.
- Adds test coverage for trusted and untrusted endpoint handling.
- Bumps the patch version.


CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsCredentialProvider.cs
```diff
@@ -202,5 +197,17 @@ public override async Task<GetAuthenticationCredentialsResponse> HandleRequestAs
             Verbose(string.Format(Resources.VSTSCredentialsNotFound, request.Uri.AbsoluteUri));
             return null;
         }
+
+        private IEnumerable<string> GetValidHosts()
+        {
+            return EnvUtil.GetHostsFromEnvironment(Logger, EnvUtil.SupportedHostsEnvVar, VstsSessionTokenClient.AllowedFeedHosts);
+        }
+
+        private static bool IsValidHost(Uri uri, IEnumerable<string> validHosts)
```
jmyersmsft: There's a mismatch here between IsValidHost, IsAllowedFeedEndpoint and IsAllowedSpsEndpoint in terms of whether the list can or cannot be overridden, and whether https is required. Also the actual matching code is repeated all 3 times.

JohnSchmeichel (approved):

CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsCredentialProvider.cs
embetten: I cleaned up the duplicate validation logic and now require HTTPS for all feed and SPS endpoints.

I kept the documented host override to avoid breaking existing custom or proxy feed scenarios. Custom hosts are now limited to the normal session-token flow and must advertise a trusted SPS endpoint. As defense in depth, the opt-in Entra-token flow ignores custom host overrides and returns tokens only to built-in trusted feed endpoints, since that flow returns the bearer token directly rather than completing the SPS exchange.

We can reconsider removing custom host overrides as a breaking change in 3.x.

jmyersmsft (approved):

